PR #1, opened 2021-01-25
Create tag images entrypoint

querti:
Entrypoint uses skopeo for copying and is able to execute the
command remotely via ssh. The sending of UMB messages has been
implemented as well.


querti:
@release-engineering/distribution review pls

pubtools/_quay/command_executor.py
```diff
@@ -0,0 +1,208 @@
...
+class Executor(object):
+    """
+    Base executor class.
+
+    Implementation of command execution should be done in
+    descendant classes. Common pre- and post-processing should be
+    implemented in this class.
+    """
+
+    def __init__(self):
```
MartinBasti, line 20: :eyes: 

why do you need abstract `__init__` here? It's already part of `object`, python stdlib

querti: You're right, I removed it.

pubtools/_quay/tag_images.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    else:
+        executor = LocalExecutor()
+
+    dest_refs = args.dest_ref if isinstance(args.dest_ref, list) else [args.dest_ref]
+    executor.skopeo_login(args.quay_user, args.quay_token)
+    executor.tag_images(args.source_ref, dest_refs)
+
+    if args.send_umb_msg:
+        from rhmsg.activemq.producer import AMQProducer
```
midnightercz, line 125: can we have this as separate method? It can be later shared with untag entry point.

emilyzheng (approved):

midnightercz (approved):

PR #9, opened 2021-03-02
- Introduced mocked entry point for testing

midnightercz:
- Introduced push-docker entry_point under "mock" group
- removed items logging
- fixed item metadata


querti:
What's the purpose of this entrypoint? To be used in pub tests? I thought we decided not to test pub's dependencies in pub. Besides, pub being able to run a mocked entrypoint doesn't mean that it will be able to run the actual one.

pubtools/_quay/push_docker.py
```diff
@@ -180,7 +162,6 @@ def push_docker(push_items, signing_key, hub, task_id, target_name, target_setti
         stepper.run(start_from=-1)
         raise
     finally:
-        log_push_items(signing_key, items=push_items)
```
lipoja, line 183: I am just curios why we do not wan to log  items?

pubtools/_quay/quay_steps.py
```diff
@@ -470,7 +470,7 @@ def _init_details(self):
         items = {}
         self._details = {"items": items}
         for item in push_items:
-            bundle_repo = list(item.metadata["destination"]["tags"].keys())[0]
+            bundle_repo = list(item.metadata["tags"].keys())[0]
```
lipoja, line 473: Is this change related to introducing mocked entry point?

midnightercz:
> I thought we decided not to test pub's dependencies in pub. Besides, pub being able to run a mocked entrypoint doesn't mean that it will be able to run the actual one.

Yes, we decided to not test pub deps inside pub. But we still don't have infrastructure which support individual dependencies and also target quay is still in intenstive implementation process. So to avoid very long testing process I thought it will be better to provide this mock entry point

pubtools/_quay/push_docker.py
midnightercz: we do want to log those, but it will require some changes in pushcollector and pub so basically it's not possible to call update_push_items here as we don't have mapping of push item states available outside of pub yet. So as log_push_items are called only in the begin and at the end of the code we can remove code from here and have it in pub instead

PR #10, opened 2021-03-04
Implement image untagging entrypoint

querti:
Entrypoint evaluates if last image references will be lost, performs the untagging and sends a UMB message.

CLOUDDST-1197


pubtools/_quay/image_untagger.py
```diff
@@ -0,0 +1,213 @@
...
+                repo_tag_mapping[repo], tag_digest_mapping, digest_tag_mapping
+            )
+            if len(lost_digests) > 0:
+                lost_imgs += [
+                    schema.format(self.host, repo, digest) for digest in lost_digests
+                ]
+
+        if not self.remove_last and lost_imgs:
+            raise ValueError(
+                "Following images would no longer be referencable by tag after"
```
emilyzheng, line 193: This sounds like images may still be reachable by digests. Is it worth also mentioning "images will be deleted when no tag references them"?

emilyzheng (approved):

pubtools/_quay/image_untagger.py
querti: I'm not really sure if the images will be actually deleted. If I uploaded a manifest list containing their digest to some tag, I think they would become referencable again. Although there's likely some garbage-collection in place.

twaugh: It's configurable at the namespace level. Generally will be e.g. 14 days.

tests/test_image_untagger.py
```diff
@@ -0,0 +1,349 @@
...
+    references = ["stage.quay.io/name/repo1:1", "stage.quay.io/name/repo2:2"]
+    token = "some-token"
+    quay_user = "some-user"
+    host = "stage.quay.io/"
+    untagger = image_untagger.ImageUntagger(
+        references, token, False, quay_user, host=host
+    )
+
+    assert untagger.references == references
+    assert untagger.host == host[:-1]
```
midnightercz, line 61: This is already tested in previous test

tests/test_image_untagger.py
```diff
@@ -0,0 +1,349 @@
...
+        "stage.quay.io/name/repo1:2",
+        "stage.quay.io/name/repo2:2",
+        "stage.quay.io/name2/repo1:1",
+        "stage.quay.io/name2/repo1:3",
+    ]
+    token = "some-token"
+    quay_user = "some-user"
+    quay_password = "some-password"
+    host = "stage.quay.io/"
+    untagger = image_untagger.ImageUntagger(
```
midnightercz, line 86: can you make this into fixture?

tests/test_image_untagger.py
```diff
@@ -0,0 +1,349 @@
...
+        expected_digest_tag_mapping = common_digest_tag_mapping
+
+        sort_dictionary_sortable_values(tag_digest_mapping)
+        sort_dictionary_sortable_values(expected_tag_digest_mapping)
+        sort_dictionary_sortable_values(digest_tag_mapping)
+        sort_dictionary_sortable_values(expected_digest_tag_mapping)
+
+        assert tag_digest_mapping == expected_tag_digest_mapping
+        assert digest_tag_mapping == expected_digest_tag_mapping
+        assert m.call_count == 3
```
midnightercz, line 142: what is the third one?

tests/test_image_untagger.py
```diff
@@ -0,0 +1,349 @@
...
+    ]
+    token = "some-token"
+    quay_user = "some-user"
+    quay_password = "some-password"
+    host = "stage.quay.io/"
+    untagger = image_untagger.ImageUntagger(
+        references, token, False, quay_user, quay_password, host
+    )
+    with requests_mock.Mocker() as m:
+        m.get(
```
midnightercz, line 255: can you make this request_mocker as fixture?

midnightercz (changes requested):

tests/test_image_untagger.py
querti: Removed.

tests/test_image_untagger.py
querti: Not sure I understand. Why can't I just invoke the class directly?

tests/test_image_untagger.py
querti: In repo_api_data fixture tags 1 and 2 are manifest lists (with the same digest). So the requests are: get repo data, get manifest list of tag 1, get manifest list of tag 2.

tests/test_image_untagger.py
querti: Hmm I don't really see how it would help.

tests/test_image_untagger.py
midnightercz: Of course you can but lines 82-88 are repeating over several tests. So to make test cases smaller it make sense to put it in fixture

tests/test_image_untagger.py
midnightercz: I see

tests/test_image_untagger.py
midnightercz: mocker setup is repeated over several tests so it can make individual test cases shorter

tests/test_image_untagger.py
querti: But the reference parameter is almost always different. How am I supposed to pass it to the fixture?

tests/test_image_untagger.py
querti: How does one make a fixture with a 'with' statement?

midnightercz: like this: https://docs.pytest.org/en/stable/fixture.html#dynamic-scope

tests/test_image_untagger.py
midnightercz: yeah, that would be ugly. So just some setup_untagger(references) function could help as well

tests/test_image_untagger.py
querti: That doesn't really answer my question, I have to use 'with' statement in the fixture itself: https://requests-mock.readthedocs.io/en/latest/mocker.html

tests/test_image_untagger.py
querti: References change in almost every test but other options changes as well sometimes. I think it will be more confusing if this setup will only be used in some specific cases instead of all

midnightercz (approved):

PR #11, opened 2021-03-11
Add authfile for skopeo login

emilyzheng:
Default authfile is ${XDG_RUNTIME_DIR}/containers/auth.json which is removed
when SSH session of a remote skopeo login terminates. $HOME/.docker/config.json
won't be removed by system logout and is checked when authorization state isn't
found in the default authfile.

### DIFF
--- a/pubtools/_quay/command_executor.py
+++ b/pubtools/_quay/command_executor.py
@@ -44,7 +44,9 @@ class Executor(object):
             )
         LOG.info("Logging in to Quay with provided credentials")
 
-        cmd_login = "skopeo login -u {0} --password-stdin quay.io".format(username)
+        cmd_login = (
+            "skopeo login --authfile $HOME/.docker/config.json -u {0} --password-stdin quay.io"
+        ).format(username)
         out, err = self._run_cmd(cmd_login, stdin=password)
 
         if "Login Succeeded" in out:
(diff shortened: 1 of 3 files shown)


querti (approved):

pubtools/_quay/command_executor.py
```diff
@@ -44,7 +44,9 @@ def skopeo_login(self, username=None, password=None):
             )
         LOG.info("Logging in to Quay with provided credentials")
 
-        cmd_login = "skopeo login -u {0} --password-stdin quay.io".format(username)
+        cmd_login = (
+            "skopeo login --authfile $HOME/.docker/config.json -u {0} --password-stdin quay.io"
+        ).format(username)
```
hluk, line 49: Look like this could allow a command injection using a constructed username.

querti: Not sure I understand, could you elaborate?

hluk: What happens if someone "accidentally" passes a username with a space?

For the command injection: What about setting username to `;rm -rf xyz;` or something?

Not sure if it's possible to do the same by overriding the `HOME` env variable (used in the command).

querti: IMO this is up to the remote host to guard against. If the user is able to supply a username/password or a private key to access the remote host, then they can simply do rm -rf directly with no injection required.

midnightercz (approved):

PR #12, opened 2021-03-22
Sanitize user input for shell commands

emilyzheng:
Convert user input to shell-escaped strings so they can be safely used in a shell command line.

### DIFF
--- a/pubtools/_quay/command_executor.py
+++ b/pubtools/_quay/command_executor.py
@@ -1,7 +1,7 @@
 import logging
-import pipes
 import shlex
 import subprocess
+from six.moves import shlex_quote
 
 import paramiko
 
@@ -47,7 +47,7 @@ class Executor(object):
 
         cmd_login = (
             "skopeo login --authfile $HOME/.docker/config.json -u {0} --password-stdin quay.io"
-        ).format(pipes.quote(username))
+        ).format(shlex_quote(username))
         out, err = self._run_cmd(cmd_login, stdin=password)
 
         if "Login Succeeded" in out:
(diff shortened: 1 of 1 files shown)


midnightercz (approved):

hluk:
+1 Nice. :)

pubtools/_quay/command_executor.py
```diff
@@ -46,7 +47,7 @@ def skopeo_login(self, username=None, password=None):
 
         cmd_login = (
             "skopeo login --authfile $HOME/.docker/config.json -u {0} --password-stdin quay.io"
-        ).format(username)
+        ).format(pipes.quote(username))
```
querti, line 50: I wonder, is this safe for Python 3? The tests passed, but the documentation makes it seem like it's deprecated. Could you pls use the 'six' module for it instead?

pubtools/_quay/command_executor.py
```diff
@@ -78,7 +79,7 @@ def tag_images(self, source_ref, dest_refs, all_arch=False):
             LOG.info(
                 "Tagging source '{0}' to destination '{1}'".format(source_ref, dest_ref)
             )
-            self._run_cmd(cmd.format(source_ref, dest_ref))
+            self._run_cmd(cmd.format(pipes.quote(source_ref), pipes.quote(dest_ref)))
```
querti, line 82: Would it be possible to move the sanitization to RemoteExecutor's _run_cmd? That way we won't have to think about it when we add new commands.

querti (changes requested):

pubtools/_quay/command_executor.py
emilyzheng: Yeah, fixed.

pubtools/_quay/command_executor.py
emilyzheng: The quote needs to be applied to user input. RemoteExecutor gets the whole command string so which part is from input can't be identified.

querti (approved):

pubtools/_quay/command_executor.py
twaugh: There's probably a more straightforward way to do this. We're quoting a string and substituting it into another string, then shlex.splitting it before passing it to subprocess.Popen. Maybe it should just be a list of strings to start with?

querti: The reason for using quote is that remote execution is done with 'paramiko' library, which doesn't support splitting the input (AFAIK)

twaugh: Ah, ok.

PR #14, opened 2021-04-16
Implement entrypoints for pub IIB methods

querti:
The new entrypoints are meant to be used by the pub methods
"PushAddIIBBundles", "PushRemoveIIBOperators", and
"PushIIBBuildFromScratch" and are invoked similiarly as push-docker.
To avoid code duplication, several methods were turned to class
methods so they could be invoked independently from the object's
context. Namely, these are: "iib_add_bundles", "iib_remove_operators",
and "run_tag_images". Image signing workflow unique to these pub
tasks was implemented in "sign_task_index_image".

One original design decision has changed by this task: it was decided
that pubtools-iib will not invoke pubtools-quay for pushing the new
index image to quay. Instead, the logic for calling pubtools-iib,
and subsequently pushing the image to its destination will be handled
in pubtools-quay. This was decided to avoid increasing the complexity
of pubtools-iib, and to simplify the implementation on pub's side.

Side effect is that the full operator workflow cannot be invoked as
a CLI script. The reason for this decision is that the original
design hasn't accounted for index image signing, which wouldn't be
performed by a CLI script anyway. The current technical solution
allows for an extension of CLI capabilities, but the argument input
would be complex and the script would still need to contact pub-hub
for the signing operation.

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -86,7 +86,7 @@ def task_iib_add_bundles(
         target_settings (dict):
             Dictionary containing settings necessary for performing the operation.
     """
-    image_schema = "{host}/{namespace}/{repo}"
+    image_schema = "{host}/{namespace}/{repo}:{tag}"
     verify_target_settings(target_settings)
 
     # Build new index image in IIB
@@ -98,19 +98,29 @@ def task_iib_add_bundles(
         target_settings=target_settings,
     )
 
-    index_image_repo = image_schema.format(
+    _, tag = build_details.index_image.split(":", 1)
+    dest_image = image_schema.format(
         host=target_settings.get("quay_host", "quay.io").rstrip("/"),
         namespace=target_settings["quay_namespace"],
         repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+        tag=tag,
     )
-    _, tag = build_details.index_image.split(":", 1)
-    dest_image = "{0}:{1}".format(index_image_repo, tag)
+    # TODO: does target_settings["quay_namespace"] point to rh-osbs?
+    # Index image used to fetch manifest list. This image will never be overwritten
+    intermediate_index_image = image_schema.format(
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo="iib",
+        tag=tag,
+    )
 
     # Sign image
     sig_handler = OperatorSignatureHandler(hub, task_id, target_settings)
-    sig_handler.sign_task_index_image(build_details, signing_key, build_details.index_image)
+    sig_handler.sign_task_index_image(signing_key, intermediate_index_image, tag)
 
     # Push image to Quay
+    # NOTE: tagging doesn't use intermediate index image, because we want the most up-to-date
+    #       image to be copied to the destination
     ContainerImagePusher.run_tag_images(
         build_details.index_image, [dest_image], True, target_settings
     )
@@ -138,7 +148,7 @@ def task_iib_remove_operators(
         target_settings (dict):
             Dictionary containing settings necessary for performing the operation.
     """
-    image_schema = "{host}/{namespace}/{repo}"
+    image_schema = "{host}/{namespace}/{repo}:{tag}"
     verify_target_settings(target_settings)
 
     # Build new index image in IIB
@@ -149,17 +159,25 @@ def task_iib_remove_operators(
         target_settings=target_settings,
     )
 
-    index_image_repo = image_schema.format(
+    _, tag = build_details.index_image.split(":", 1)
+    dest_image = image_schema.format(
         host=target_settings.get("quay_host", "quay.io").rstrip("/"),
         namespace=target_settings["quay_namespace"],
         repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+        tag=tag,
     )
-    _, tag = build_details.index_image.split(":", 1)
-    dest_image = "{0}:{1}".format(index_image_repo, tag)
+
+    # Index image used to fetch manifest list. This image will never be overwritten
+    intermediate_index_image = image_schema.format(
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo="iib",
+        tag=tag,
+    )
 
     # Sign image
     sig_handler = OperatorSignatureHandler(hub, task_id, target_settings)
-    sig_handler.sign_task_index_image(build_details, signing_key, build_details.index_image)
+    sig_handler.sign_task_index_image(signing_key, intermediate_index_image, tag)
 
     # Push image to Quay
     ContainerImagePusher.run_tag_images(
@@ -189,7 +207,7 @@ def task_iib_build_from_scratch(
         target_settings (dict):
             Dictionary containing settings necessary for performing the operation.
     """
-    image_schema = "{host}/{namespace}/{repo}"
+    image_schema = "{host}/{namespace}/{repo}:{tag}"
     verify_target_settings(target_settings)
 
     # Build new index image in IIB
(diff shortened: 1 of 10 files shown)


pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,245 @@
...
+        target_settings=target_settings,
+    )
+
+    # Push image to Quay
+    index_image_repo = image_schema.format(
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+    )
+    # TODO: how is tag set? Do we derive it from source or destination index image?
```
querti, line 107: @midnightercz

midnightercz (changes requested):

pubtools/_quay/operator_pusher.py
```diff
@@ -215,46 +249,32 @@ def iib_add_bundles(self, bundles, archs, ocp_version, deprecation_list=None):
...
         # inconsistent way of presenting multiple arguments...
         if deprecation_list:
-            args += ["--deprecation-list", ",".join(deprecation_list)]
-
-        env_vars = {}
-        if "iib_overwrite_from_index_token" in self.target_settings:
-            env_vars["OVERWRITE_FROM_INDEX_TOKEN"] = self.target_settings[
-                "iib_overwrite_from_index_token"
-            ]
+            args += ["--deprecation-list", deprecation_list]
```
emilyzheng, line 277: This argument expects a string.

querti: The value is taken from pub. Pub doesn't parse it so it's only a comma separated string.

emilyzheng: It's taken from pub for iib-add-bundles entrypoint. But for push_docker, it's from get_deprecation_list which returns a list.
```
            deprecation_list ([str]):
                List of bundles to be deprecated.
```
This is also incorrect if it's expected to be a string.

pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,245 @@
...
+    _, tag = build_details.index_image.split(":", 1)
+    dest_image = "{0}:{1}".format(index_image_repo, tag)
+
+    ContainerImagePusher.run_tag_images(
+        build_details.index_image, [dest_image], True, target_settings
+    )
+
+    # Sign image
```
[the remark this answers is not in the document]
querti: Changed.

pubtools/_quay/operator_pusher.py
querti: You're right, fixed.

pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,243 @@
...
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+    )
+    _, tag = build_details.index_image.split(":", 1)
+    dest_image = "{0}:{1}".format(index_image_repo, tag)
+
+    # Sign image
+    sig_handler = OperatorSignatureHandler(hub, task_id, target_settings)
+    sig_handler.sign_task_index_image(build_details, signing_key, dest_image)
```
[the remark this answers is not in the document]
querti: Great catch, fixed

pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,245 @@
...
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+    )
+    _, tag = build_details.index_image.split(":", 1)
+    dest_image = "{0}:{1}".format(index_image_repo, tag)
+
+    # Sign image
+    sig_handler = OperatorSignatureHandler(hub, task_id, target_settings)
+    sig_handler.sign_task_index_image(build_details, signing_key, build_details.index_image)
```
emilyzheng, line 111: I think build_details.index_image_resolved is the right one to be signed since build_details.index_image could be overwritten.

pubtools/_quay/signature_handler.py
```diff
@@ -580,10 +580,44 @@ def sign_operator_images(self, iib_results):
...
+
+        self.upload_signatures_to_pyxis(
+            claim_messages,
+            signature_messages,
+            self.target_settings.get(
+                "sigstore_max_upload_items", self.DEFAULT_MAX_ITEMS_PER_UPLOAD_BATCH
+            ),
+        )
+
+    def sign_task_index_image(self, build_details, signing_key, index_image, tag=None):
```
emilyzheng, line 599: Seems parameter 'index_image' is unnecessary if both index_image and tag can be got from build_details. And build_details is unused in this function.

pubtools/_quay/iib_operations.py
querti: From what I understand of the hello-operator guild discussion, new index image is always built as iib:<task_id>. So I guess I'll have to change the image name to "iib" to ensure that stable image is used. @midnightercz can you confirm or (if I'm wrong) explain the proper procedure?

pubtools/_quay/signature_handler.py
querti: You're right, but if the index image name has to be changed to "iib", it would probably be preferable to keep the "index_image" argument and remove "build_details". I'll wait for confirmation from @midnightercz

pubtools/_quay/iib_operations.py
querti: I changed the implementation to use intermediate index image

pubtools/_quay/signature_handler.py
querti: Removed the redundant parameter

pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,270 @@
...
+        namespace=target_settings["quay_namespace"],
+        repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+        tag=tag,
+    )
+    # TODO: does target_settings["quay_namespace"] point to rh-osbs?
+    # Index image used to fetch manifest list. This image will never be overwritten
+    intermediate_index_image = image_schema.format(
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo="iib",
```
midnightercz, line 113: This is not true for staging iib (it's iib-stage there). Emily pointed out in my code review in pub and it applies also here.

pubtools/_quay/iib_operations.py
```diff
@@ -0,0 +1,270 @@
...
+        namespace=target_settings["quay_namespace"],
+        repo=get_internal_container_repo_name(target_settings["quay_operator_repository"]),
+        tag=tag,
+    )
+
+    # Index image used to fetch manifest list. This image will never be overwritten
+    intermediate_index_image = image_schema.format(
+        host=target_settings.get("quay_host", "quay.io").rstrip("/"),
+        namespace=target_settings["quay_namespace"],
+        repo="iib",
```
midnightercz, line 174: This is not true for staging iib (it's iib-stage there). Emily pointed out in my code review in pub and it applies also here.

midnightercz (changes requested):

pubtools/_quay/iib_operations.py
midnightercz: Actually I was wrong, it's really iib. Only org is different per envs, which is correct here

pubtools/_quay/iib_operations.py
midnightercz: Actually I was wrong, it's really iib. Only org is different per envs, which is correct here

midnightercz (approved):

emilyzheng (approved):

PR #20, opened 2021-05-12
Support external repos for push items

midnightercz:
When push items are populated by ET, they include both repos
and external_repos. In the case external repos are set, they
should be prioritized and used instead of 'repos'

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -231,7 +231,6 @@ class PushDocker:
         """
         repos = []
         for item in push_items:
-            LOG.info("ITEM %s", item)
             if item.external_repos:
                 repos += item.external_repos.keys()
             else:
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -150,7 +150,7 @@ class PushDocker:
         for push_item in push_items:
             if push_item.external_repos:
                 #  if external repos is defined, push items was populated by ET
-                #  item.repos have to be remove from tags mapping
+                #  item.repos have to be removed from tags mapping
                 for repo in list(push_item.repos):
                     push_item.metadata["tags"].pop(repo)
 
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -3,16 +3,8 @@ import logging
 
 import requests
 
-from .exceptions import (
-    BadPushItem,
-    InvalidTargetSettings,
-    InvalidRepository,
-)
-from .utils.misc import (
-    run_entrypoint,
-    get_internal_container_repo_name,
-    log_step,
-)
+from .exceptions import BadPushItem, InvalidTargetSettings, InvalidRepository
+from .utils.misc import run_entrypoint, get_internal_container_repo_name, log_step
 from .quay_api_client import QuayApiClient
 from .quay_client import QuayClient
 from .container_image_pusher import ContainerImagePusher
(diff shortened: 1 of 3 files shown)


pubtools/_quay/push_docker.py
```diff
@@ -229,7 +231,12 @@ def check_repos_validity(self, push_items):
         """
         repos = []
         for item in push_items:
-            repos += item.repos.keys()
+            LOG.info("ITEM %s", item)
```
emilyzheng, line 234: Debugging log?

emilyzheng (approved):

pubtools/_quay/push_docker.py
midnightercz: right, I need to remove this

pubtools/_quay/push_docker.py
```diff
@@ -152,6 +144,16 @@ def get_docker_push_items(self):
 
         return docker_push_items
 
+    @log_step("Translate docker push items")
+    def filter_unrelated_repos(self, push_items):
+        """Remove item repos from tag mapping if external_repos is set for push item."""
```
querti, line 149: Could you write a full docstring including the arguments?

pubtools/_quay/push_docker.py
```diff
@@ -152,6 +144,16 @@ def get_docker_push_items(self):
...
+        for push_item in push_items:
+            if push_item.external_repos:
+                #  if external repos is defined, push items was populated by ET
+                #  item.repos have to be remove from tags mapping
```
querti, line 153:
```suggestion
                #  item.repos have to be removed from tags mapping
```

tests/test_push_docker.py
```diff
@@ -390,14 +399,18 @@ def test_check_repos_validity_success(
...
+            container_push_item_correct_repos,
+            container_signing_push_item,
+        ]
     )
 
     mock_get_target_info.assert_called_once_with("target_stage_quay")
-    assert mock_get_repo_metadata.call_count == 2
+    assert mock_get_repo_metadata.call_count == 3
     mock_get_repo_metadata.call_args_list[0] == mock.call("namespace/repo1")
     mock_get_repo_metadata.call_args_list[1] == mock.call("namespace/repo2")
```
querti, line 412: Can you add checks for the extra calls?

tests/test_push_docker.py
```diff
@@ -830,21 +850,28 @@ def test_push_docker_full_success(
 
     mock_get_docker_push_items.assert_called_once_with()
     mock_get_docker_push_items.assert_called_once_with()
-    mock_check_repos_validity.assert_called_once_with([container_multiarch_push_item])
-    mock_generate_backup_mapping.assert_called_once_with([container_multiarch_push_item])
+    mock_check_repos_validity.assert_called_once_with(
```
querti, line 853: Could you add a check if filter_unrealted_repos was called?

pubtools/_quay/push_docker.py
```diff
@@ -152,6 +144,16 @@ def get_docker_push_items(self):
 
         return docker_push_items
 
+    @log_step("Translate docker push items")
+    def filter_unrelated_repos(self, push_items):
```
querti, line 148: Could you add a separate test for this new method?

querti (changes requested):

tests/test_push_docker.py
midnightercz: it's not mocked so it has to be called

querti (approved):

PR #21, opened 2021-05-18
Add tag checks to tag-docker

querti:
Two checks are added:
1. If adding a new tag to production target, this tag must already
exist in stage target.
2. If removing a tag in a production target, this tag must already
not exist in stage target.

### DIFF
--- a/pubtools/_quay/tag_docker.py
+++ b/pubtools/_quay/tag_docker.py
@@ -59,15 +59,15 @@ class TagDocker:
         self.target_name = target_name
         self.target_settings = target_settings
 
-        self.verify_target_settings()
-        self.verify_input_data()
-
-        self.quay_host = self.target_settings.get("quay_host", "quay.io").rstrip("/")
-
         # TODO: will our robot credentials be able to read from brew's build repos?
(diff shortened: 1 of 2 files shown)


querti:
Refers to CLOUDDST-402

pubtools/_quay/tag_docker.py
```diff
@@ -150,6 +150,50 @@ def verify_input_data(self):
             if item.metadata["tag_source"] and ":" in item.metadata["tag_source"]:
                 raise BadPushItem("Specifying source via digest is not allowed.")
 
+            # if prod target is selected, additional checks are made
```
midnightercz, line 153: Can you please move this to another method which isn't called called in the __init__() but in run()? I don't think it's good idea to start querying quay immediately when instance is initiated

midnightercz (changes requested):

pubtools/_quay/tag_docker.py
querti: Fixed

emilyzheng (approved):

midnightercz (approved):

PR #23, opened 2021-05-20
Remove signature of outdated containers in push-docker operation.

midnightercz:
When push docker operation overwrites floating tags, there's
no need to keep signatures for them and therefore they can
be removed.


pubtools/_quay/push_docker.py
```diff
@@ -422,6 +417,16 @@ def run(self):
             LOG.error("An exception has occurred during the push, starting rollback")
             self.rollback(backup_tags, rollback_tags)
             raise e
+        else:
+            # Remove old signatures
+            container_signature_handler.remove_outdated_signatures(
+                [
+                    (image_data.repo, manifest["digest"], image_data.tag)
```
querti, line 424: Are you getting these digests correctly? The manifest will be either manifest list in which case you need to iterate over "manifests" to get all digests. Or (in case of source image) it will be a V2S2 manifest which doesn't contain digest data. You may use the same approach as "generate_backup_mapping" to gather this digest.

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +368,37 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
...
+            repo, digest, reference = sig_to_remove
+            args = [
+                "--pyxis-server",
+                self.target_settings["pyxis_server"],
+                "--pyxis-krb-principal",
+                self.target_settings["iib_krb_principal"],
+            ]
+            if "iib_krb_ktfile" in self.target_settings:
+                args += ["--pyxis-krb-ktfile", self.target_settings["iib_krb_ktfile"]]
+            args += ["--repo", repo, "--digest", digest, "--reference", reference]
```
querti, line 391: Could you check PR https://github.com/release-engineering/pubtools-pyxis/pull/20/files ? The signatures search supports the same inputs as get_signatures entrypoint. I would suggest using just digest to match to-be-removed signatures.

Although now that I think about it more, it's possible that an image with the same digest was reuploaded, generating no new signatures. In this case it would be undesirable to remove these signatures.

Perhaps the right approach is to get existing signatures by digest (using https://github.com/release-engineering/pubtools-pyxis/pull/18), check them against the result of construct_item_claim_messages, and only remove those which were not re-added. To-be-deleted signatures would be referenced by pyxis's internal ID. This means that I would have to change my PR https://github.com/release-engineering/pubtools-pyxis/pull/20 to accept signature IDs

Have I missed something? What do you think?

querti (changes requested):

pubtools/_quay/signature_handler.py
midnightercz: I think if I simply remove only signature which are not in manifest_claims it should be enough. And combination of manifest digest + reference should be unique enough.

pubtools/_quay/signature_handler.py
```diff
@@ -410,6 +443,27 @@ def validate_radas_messages(self, claim_messages, signature_messages):
...
+        """
+        Remove outdated signatures from sigstore.
+
+        Method removes signatures for all referenced registries
+
+        Args:
+            outdated_containers [(repo, manifest, tag)]:
+                List of tuples container repo, manifest and tag of outdated container
+
+
```
querti, line 456: Can you remove the whitespaces?

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +415,56 @@ def run(self):
...
+            # Remove old signatures
+
+            new_signatures = [
+                (m["manifest_digest"], m["docker_reference"].split(":")[1]) for m in manifest_claims
+            ]
+            outdated_signatures = []
+            for image_data, manifest in backup_tags.items():
+                if "manifests" in manifest:
+                    for arch_manifest in manifest["manifests"]:
+                        if (arch_manifest["digest"], image_data.tag) not in new_signatures:
```
querti, line 455: Is this a sufficient comparison method? Shouldn't repo be considered as well?

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +415,56 @@ def run(self):
...
             raise e
+        else:
+            # Remove old signatures
+
+            new_signatures = [
+                (m["manifest_digest"], m["docker_reference"].split(":")[1]) for m in manifest_claims
+            ]
+            outdated_signatures = []
+            for image_data, manifest in backup_tags.items():
+                if "manifests" in manifest:
```
querti, line 453: Are source images not considered?

pubtools/_quay/signature_handler.py
```diff
@@ -611,6 +666,7 @@ def sign_operator_images(self, iib_results):
                 "sigstore_max_upload_items", self.DEFAULT_MAX_ITEMS_PER_UPLOAD_BATCH
             ),
         )
+        return (claim_messages, signature_messages)
```
querti, line 669: Here as well

querti (changes requested):

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +415,56 @@ def run(self):
...
+                    for arch_manifest in manifest["manifests"]:
+                        if (arch_manifest["digest"], image_data.tag) not in new_signatures:
+                            outdated_signatures.append((arch_manifest, image_data.tag))
```
querti, line 456: remove_outdated_signatures requires repo, manifest, tag, no?

querti (changes requested):

querti (changes requested):
Personally, I think the best solution will be to use pubtools-pyxis-get-signatures to get removal candidate signatures, and filter out those that would be removed. Afterwards, delete-signature entrypoint will only be called with Pyxis internal IDs of signatures as arguments.

What do you think?

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +368,49 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
                 env_vars,
             )
 
+    def remove_signatures_from_pyxis(self, signatures_to_remove):
+        """
+        Upload signatures to Pyxis by using a pubtools-pyxis entrypoint.
+
+        Args:
+            signature_to_remove ([repo, tag, manifest]):
```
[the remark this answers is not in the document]
midnightercz: Right, original method work with that. Then I rewrote it

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +415,56 @@ def run(self):
...
+            )
             # Push container images
             container_pusher = ContainerImagePusher(docker_push_items, self.target_settings)
             container_pusher.push_container_images()
 
             if operator_push_items:
                 # Build index images
                 operator_pusher = OperatorPusher(operator_push_items, self.target_settings)
+                existing_index_images = operator_pusher.get_existing_index_images(
+                    self.quay_api_client
```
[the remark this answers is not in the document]
midnightercz: It's only used for this one specific call. I didn't want to bind those classes together more then it's necessary

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +416,77 @@ def run(self):
...
+                    for arch_manifest in manifest["manifests"]:
+                        outdated_signatures.append((arch_manifest, image_data.tag, ext_repo))
+                else:
+                    outdated_signatures.append((manifest["digest"], image_data.tag, ext_repo))
+
+            signatures_to_remove = []
+            for esig in container_signature_handler.get_signatures_from_pyxis(
+                [sig[0] for sig in outdated_signatures]
+            ):
+                if (
```
querti, line 466: Can you explain the purpose of this check? To see if signatures that should exist actually exist? It would make sense if we actually mapped unique signature IDs to the required fields

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +416,77 @@ def run(self):
...
+            container_signature_handler.remove_outdated_signatures(signatures_to_remove)
+
+            signatures_to_remove = []
+            if operator_push_items:
+                outdated_signatures = []
+                new_operator_signatures = [
+                    (m["manifest_digest"], m["docker_reference"]) for m in manifest_claims
+                ]
+                for esig in container_signature_handler.get_signatures_from_pyxis(
+                    [repo_digest_version[1] for repo_digest_version in existing_index_images]
```
querti, line 485: Aren't digests the first element of the tuple, not second?

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +416,77 @@ def run(self):
...
+
+            container_signature_handler.remove_outdated_signatures(signatures_to_remove)
+
+            signatures_to_remove = []
+            if operator_push_items:
+                outdated_signatures = []
+                new_operator_signatures = [
+                    (m["manifest_digest"], m["docker_reference"]) for m in manifest_claims
+                ]
+                for esig in container_signature_handler.get_signatures_from_pyxis(
```
querti, line 484: Unless we decide to use signature IDs I feel like there's no need to actually gather existing signatures

querti (changes requested):
Could you also address my previous comments?

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +416,77 @@ def run(self):
...
+                    iib_results
+                )
                 # Push index images to Quay
                 operator_pusher.push_index_images(iib_results)
         except Exception as e:
             LOG.error("An exception has occurred during the push, starting rollback")
             self.rollback(backup_tags, rollback_tags)
             raise e
+        else:
+            # Remove old signatures
```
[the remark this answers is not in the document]
midnightercz: sure

pubtools/_quay/push_docker.py
```diff
@@ -446,25 +447,46 @@ def run(self):
...
             outdated_signatures = []
+
             for image_data, manifest in backup_tags.items():
+                ext_repo = image_data.repo.split("/")[1].replace(INTERNAL_DELIMITER, "/")
                 if "manifests" in manifest:
                     for arch_manifest in manifest["manifests"]:
-                        if (arch_manifest["digest"], image_data.tag) not in new_signatures:
-                            outdated_signatures.append((arch_manifest, image_data.tag))
-            container_signature_handler.remove_outdated_signatures(outdated_signatures)
+                        outdated_signatures.append((arch_manifest, image_data.tag, ext_repo))
```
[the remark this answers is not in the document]
midnightercz: right

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +368,49 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
...
+            "--pyxis-server",
+            self.target_settings["pyxis_server"],
+            "--pyxis-krb-principal",
+            self.target_settings["iib_krb_principal"],
+        ]
+        if "iib_krb_ktfile" in self.target_settings:
+            args += ["--pyxis-krb-ktfile", self.target_settings["iib_krb_ktfile"]]
+        args += [
+            "--digest",
+            ",".join(digests).rstrip(","),
```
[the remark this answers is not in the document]
midnightercz: sure, if you do counterpart change in pubtools-pyxis

pubtools/_quay/push_docker.py
midnightercz: Main reasons for that are: 1. avoid removal of non existing signatures, remove also signature for reference registries which doesn't have to be in target settings (because they were removed for example).

pubtools/_quay/push_docker.py
midnightercz: existing signatures are fetched only by digests, so in return could be also signatures referring another repo (or tags). And second part of course checks if existing signatures isn't in "new" signatures which meant to be added in the current push

pubtools/_quay/push_docker.py
```diff
@@ -415,26 +416,77 @@ def run(self):
...
+            ]
+            outdated_signatures = []
+
+            for image_data, manifest in backup_tags.items():
+                ext_repo = image_data.repo.split("/")[1].replace(INTERNAL_DELIMITER, "/")
+                if "manifests" in manifest:
+                    for arch_manifest in manifest["manifests"]:
+                        outdated_signatures.append((arch_manifest, image_data.tag, ext_repo))
+                else:
+                    outdated_signatures.append((manifest["digest"], image_data.tag, ext_repo))
```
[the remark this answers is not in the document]
midnightercz: digest is from manifest in backup tags

pubtools/_quay/signature_handler.py
```diff
@@ -410,6 +432,20 @@ def validate_radas_messages(self, claim_messages, signature_messages):
                 )
             )
 
+    def remove_outdated_signatures(self, signatures_to_remove):
```
querti, line 435: Is this method even necessary?

pubtools/_quay/signature_handler.py
```diff
@@ -518,6 +572,7 @@ def sign_container_images(self, push_items):
                 "sigstore_max_upload_items", self.DEFAULT_MAX_ITEMS_PER_UPLOAD_BATCH
             ),
         )
+        return (claim_messages, signature_messages)
```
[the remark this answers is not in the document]
querti: bump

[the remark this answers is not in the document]
querti: bump

pubtools/_quay/signature_handler.py
querti: Could you only return claim messages? Could you also update the docstring?

pubtools/_quay/signature_handler.py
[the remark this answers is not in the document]
querti: Could you also update the docstring?

pubtools/_quay/operator_pusher.py
```diff
@@ -326,6 +324,39 @@ def iib_remove_operators(cls, operators=None, archs=None, index_image=None, targ
...
+            image_ref = "{0}:{1}".format(index_image_repo, version)
+            try:
+                manifest_list = quay_client.get_manifest(image_ref, manifest_list=True)
+            except requests.exceptions.HTTPError as e:
+                if e.response.status_code == 404:
+                    continue
+                else:
+                    raise
+            for manifest in manifest_list["manifests"]:
+                current_index_images.append((manifest["digest"], version))
```
querti, line 357: It seems like the tag is not used anywhere. Is there a need to return it?

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +369,37 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
                 env_vars,
             )
 
+    def remove_signatures_from_pyxis(self, signatures_to_remove):
+        """
+        Remove signatures from Pyxis by using a pubtools-pyxis entrypoint.
+
+        Args:
+            signatures_to_remove (list):
```
querti, line 377:
```suggestion
            signatures_to_remove ([str]):
```

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +369,37 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
...
+                list of signature ids to be removed
```
querti, line 378:
```suggestion
                List of signature ids to be removed.
```

pubtools/_quay/signature_handler.py
```diff
@@ -378,6 +369,37 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
...
+        """
+        LOG.info("Removing outdated signatures from pyxis")
```
querti, line 380:
```suggestion
        LOG.info("Removing outdated signatures from Pyxis")
```

querti (changes requested):

pubtools/_quay/operator_pusher.py
```diff
@@ -326,6 +324,39 @@ def iib_remove_operators(cls, operators=None, archs=None, index_image=None, targ
...
+        image_schema = "{host}/{namespace}/{repo}"
+        iib_repo = get_internal_container_repo_name(
+            self.target_settings["quay_operator_repository"]
+        )
+        index_image_repo = image_schema.format(
+            host=self.quay_host, namespace=self.target_settings["quay_namespace"], repo=iib_repo
+        )
+        current_index_images = []
+
+        manifest_list = {}
```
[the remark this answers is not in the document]
midnightercz: yeah, this one can be removed

pubtools/_quay/signature_handler.py
midnightercz: yes, I guess it doesn't make sense. There won't be any extra logic needed for this in the future

pubtools/_quay/operator_pusher.py
midnightercz: It's not necessary to return it for purpose of cleaning up outdated signatures

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +378,54 @@ def rollback(self, backup_tags, rollback_tags):
...
+        new_signatures = [(m["manifest_digest"], m["docker_reference"]) for m in manifest_claims]
+        outdated_signatures = []
+
+        for image_data, manifest in backup_tags.items():
+            ext_repo = image_data.repo.split("/")[1].replace(INTERNAL_DELIMITER, "/")
+            if "manifests" in manifest:
+                for arch_manifest in manifest["manifests"]:
+                    outdated_signatures.append((arch_manifest["digest"], image_data.tag, ext_repo))
+            else:
+                outdated_signatures.append((manifest["digest"], image_data.tag, ext_repo))
```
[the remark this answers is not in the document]
midnightercz: fixed

pubtools/_quay/push_docker.py
```diff
@@ -331,11 +332,13 @@ def generate_backup_mapping(self, push_items):
                 for tag in tags:
                     # repo doesn't exist, add to rollback tags
                     if not repo_data:
-                        rollback_tags.append(PushDocker.ImageData(full_repo, tag))
+                        rollback_tags.append(PushDocker.ImageData(full_repo, tag, None))
```
querti, line 335: Could you explain this None in the docstring? For a new reader it would be unclear why this is done, and that this structure is used for signature removal as well.

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +380,89 @@ def rollback(self, backup_tags, rollback_tags):
...
+                esig["manifest_digest"],
+                esig["reference"],
+            ) not in new_signatures:
+                signatures_to_remove.append(esig["_id"])
+
+        container_signature_handler.remove_signatures_from_pyxis(signatures_to_remove)
+
+        for item in push_items:
+            claim_messages += container_signature_handler.construct_item_claim_messages(item)
+        new_signatures = [(m["manifest_digest"], m["docker_reference"]) for m in claim_messages]
```
querti, line 446: This whole part shouldn't be here, right?

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +380,89 @@ def rollback(self, backup_tags, rollback_tags):
...
+                operator_claim_messages += (
+                    container_signature_handler.construct_item_claim_messages(item)
+                )
+            new_operator_signatures = [
+                (m["manifest_digest"], m["docker_reference"]) for m in operator_claim_messages
+            ]
+            for esig in container_signature_handler.get_signatures_from_pyxis(
+                [digest_version[0] for digest_version in existing_index_images]
+            ):
+                if (esig["manifest_digest"], esig["reference"]) not in new_operator_signatures:
```
querti, line 462: Is it even possible for new signatures to have old manifest digest?

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +380,89 @@ def rollback(self, backup_tags, rollback_tags):
...
+            ) not in new_signatures:
+                signatures_to_remove.append(esig["_id"])
+
+        container_signature_handler.remove_signatures_from_pyxis(signatures_to_remove)
+
+        for item in push_items:
+            claim_messages += container_signature_handler.construct_item_claim_messages(item)
+        new_signatures = [(m["manifest_digest"], m["docker_reference"]) for m in claim_messages]
+
+        signatures_to_remove = []
```
querti, line 448: Is it possible that and index image would be signed for some other tag? Do we need to check that the references correspond with each other?

querti (changes requested):

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +380,89 @@ def rollback(self, backup_tags, rollback_tags):
...
+                before new index image was pushed in the current task
+            backup_tags ({ImageData: str}):
+                Dictionary of ImageData (repo, tag, digest) -> manifest
+                holding containers which were overwritten in the currently running task
+            container_signature_handler (ContainerSignatureHandler):
+                ContanerSignatureHandler instance.
+        """
+        claim_messages = []
+        for item in push_items:
+            claim_messages += container_signature_handler.construct_item_claim_messages(item)
```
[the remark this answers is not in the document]
midnightercz: I found it strange to pass manifests claims as argument but in sign_container_images call filter on those. Or do you prefer to call filter also here in PushDocker.run method?

pubtools/_quay/push_docker.py
midnightercz: I'll put here comments explaining why there's sometimes None and sometimes digest

pubtools/_quay/push_docker.py
midnightercz: yeah, copy paste accident

pubtools/_quay/push_docker.py
midnightercz: In theory it is. In practice, hard to tell. Maybe opm also does modifications which includes timestamps

pubtools/_quay/push_docker.py
midnightercz: I guess it's very improbable, but sure I can rather add check here as well

pubtools/_quay/push_docker.py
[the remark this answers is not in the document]
querti: I think you can still return them from sign_container_images method, just save them somewhere after 'remove_duplicate_claim_messages' and before 'filter_claim_messages'.

querti (approved):

pubtools/_quay/push_docker.py
```diff
@@ -293,7 +294,7 @@ def generate_backup_mapping(self, push_items):
         """
         Create resources which will be used for rollback if something goes wrong during the push.
 
-        Specifically, create two resources: 'backup_tags' and 'rollback_tags'.
+        Specifically, create two resources: 'backup_tags' and 'pollback_tags'.
```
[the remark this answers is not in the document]
midnightercz: right, probably vim accident

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +383,90 @@ def rollback(self, backup_tags, rollback_tags):
...
+                esig["manifest_digest"],
+                esig["reference"].split(":")[-1],
+                esig["repository"],
+            ) in outdated_signatures and (
+                esig["manifest_digest"],
+                esig["reference"],
+            ) not in new_signatures:
+                signatures_to_remove.append(esig["_id"])
+
+        container_signature_handler.remove_signatures_from_pyxis(signatures_to_remove)
```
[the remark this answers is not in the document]
midnightercz: yeah, it can be simplified

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +383,90 @@ def rollback(self, backup_tags, rollback_tags):
...
+
+        signatures_to_remove = []
+        if existing_index_images:
+            outdated_signatures = []
+            operator_claim_messages = []
+            for item in operator_push_items:
+                operator_claim_messages += (
+                    container_signature_handler.construct_item_claim_messages(item)
```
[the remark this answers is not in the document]
midnightercz: fixed

pubtools/_quay/push_docker.py
```diff
@@ -443,19 +443,25 @@ def remove_old_signatures(
...
+            image_schema = "{host}/{repository}:{tag}"
+            new_operator_signatures = []
+            for dest_reg in self.target_settings["docker_settings"]["docker_reference_registry"]:
+                new_operator_signatures.extend(
+                    [
+                        (
+                            eii[0],
+                            image_schema.format(host=dest_reg, repository=iib_repo, tag=eii[1]),
+                        )
+                        for eii in existing_index_images
```
emilyzheng, line 462: existing_index_images are images before iib-add-bundles. Shouldn't new signatures be got from images after iib operation?

midnightercz: right, I mixed that up. Fixed

pubtools/_quay/signature_handler.py
```diff
@@ -381,6 +372,37 @@ def upload_signatures_to_pyxis(self, claim_mesages, signature_messages, max_item
                 env_vars,
             )
 
+    def remove_signatures_from_pyxis(self, signatures_to_remove):
```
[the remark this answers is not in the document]
midnightercz: I understood how you meant it. I removed the duplicate method

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +384,119 @@ def rollback(self, backup_tags, rollback_tags):
...
+                [digest_version[0] for digest_version in existing_index_images]
+            ):
+                if (esig["manifest_digest"], esig["reference"]) not in new_operator_signatures and (
+                    esig["manifest_digest"],
+                    esig["reference"].split(":")[-1],
+                    esig["repository"],
+                ) in existing_index_images:
+                    signatures_to_remove.append(esig["_id"])
+            if signatures_to_remove:
+                signature_remover.remove_signatures_from_pyxis(signatures_to_remove)
```
emilyzheng, line 498: pyxis parameters are missing.

pubtools/_quay/push_docker.py
```diff
@@ -377,6 +384,119 @@ def rollback(self, backup_tags, rollback_tags):
...
+                esig["reference"],
+            ) not in new_signatures:
+                signatures_to_remove.append(esig["_id"])
+
+        if signatures_to_remove:
+            signature_remover.remove_signatures_from_pyxis(
+                signatures_to_remove,
+                self.target_settings["pyxis_server"],
+                self.target_settings["pyxis_krb_principal"],
+                self.target_settings["pyxis_krb_ktfile"],
```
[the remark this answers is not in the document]
midnightercz: Hmm, you're right. That's probably something we should fix in the future

emilyzheng (approved):

PR #25, opened 2021-05-25
Skip signing when signing key is None

emilyzheng:
Container signing key is got from claims_signing_key of push items.
Value of claims_signing_key could be None if it's not provided by
user. Skip creating claim messages for key None so signing won't
happen for it.

### DIFF
--- a/tests/test_signature_handler.py
+++ b/tests/test_signature_handler.py
@@ -800,3 +800,45 @@ def test_sign_claim_messages_no_signatures(
     mock_get_radas_signatures.assert_not_called()
     mock_validate_radas_msgs.assert_not_called()
     mock_upload_signatures_to_pyxis.assert_not_called()
+
+
+@mock.patch("pubtools._quay.signature_handler.QuayClient")
+@mock.patch("pubtools._quay.signature_handler.QuayApiClient")
+def test_construct_item_claim_messages_none_signing_key(
+    mock_quay_api_client,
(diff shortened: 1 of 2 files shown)


tests/test_signature_handler.py
```diff
@@ -800,3 +800,45 @@ def test_sign_claim_messages_no_signatures(
     mock_get_radas_signatures.assert_not_called()
     mock_validate_radas_msgs.assert_not_called()
     mock_upload_signatures_to_pyxis.assert_not_called()
+
+
+@mock.patch("pubtools._quay.signature_handler.QuayClient")
+@mock.patch("pubtools._quay.signature_handler.QuayApiClient")
+def test_construct_item_claim_messages_invalid_signing_key(
```
midnightercz, line 807: can we please rename this to on_signing_key. In the future we'll implement condition to accept only 16 characters long signing key and therefore invalid signing key will be anything shorter. So current name could lead to misunderstanding

midnightercz (changes requested):

tests/test_signature_handler.py
emilyzheng: Yeah, updated.

querti (approved):
I think similar logic needs to be added to tag-docker. It could be a separate PR, since tag-docker is not verified yet.

chandwanitulsi (approved):

PR #31, opened 2021-06-07
Add entrypoint for clearing a Quay repo 

querti:
The code for removing the signatures of untagged images is taken
from repo removjng entrypoint, as the same principle applies.

### DIFF
--- a/pubtools/_quay/clear_repo.py
+++ b/pubtools/_quay/clear_repo.py
@@ -20,8 +20,8 @@ CLEAR_REPO_ARGS = {
         "required": True,
         "type": str,
     },
-    ("--namespace",): {
-        "help": "Internal Quay namespace in which repository resides.",
+    ("--quay-org",): {
+        "help": "Quay organization in which repository resides.",
         "required": True,
         "type": str,
     },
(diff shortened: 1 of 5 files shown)


emilyzheng (approved):

pubtools/_quay/clear_repo.py
```diff
@@ -0,0 +1,256 @@
...
+logging.basicConfig()
+LOG.setLevel(logging.INFO)
+
+CLEAR_REPO_ARGS = {
+    ("--repository",): {
+        "help": "External repository to clear. Must be in format <namespace>/<repo>.",
+        "required": True,
+        "type": str,
+    },
+    ("--namespace",): {
```
midnightercz, line 23: can we rename it to org or quay-org? namespace could be misleading because it's already part of repo

midnightercz (changes requested):

pubtools/_quay/clear_repo.py
querti: Changed. I assumed you also wanted it changed for the remove-repo entrypoint

midnightercz (approved):

PR #33, opened 2021-06-14
refactor: drop unnecessary 'external_repos'

rohanpm:
Pub refactor will make it so that the preferred repo format is always
used in 'repos' attribute. We no longer need to check two places for
repos or rewrite other attributes of our push items.

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -146,21 +146,6 @@ class PushDocker:
 
         return docker_push_items
 
-    @log_step("Translate docker push items")
-    def filter_unrelated_repos(self, push_items):
-        """Remove item repos from tag mapping if external_repos is set for push item.
-
-        Args:
-            push_items ([ContainerPushItem]):
-                Container push items containing the repositories.
-        """
-        for push_item in push_items:
-            if push_item.external_repos:
-                #  if external repos is defined, push items was populated by ET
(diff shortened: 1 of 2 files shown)


pubtools/_quay/push_docker.py
```diff
@@ -439,10 +422,7 @@ def run(self):
         # Return repos for UD cache flush
```
rohanpm, line 422: I don't quite understand this: I checked where we use the UD cache flush API in Pub and it's the same as we do when flushing rhsm-pulp repos.

So, given that API expects rhsm-pulp repo IDs, why would we ask it to do cache flushes for either of:

- docker-pulp repo IDs (note: this case will be removed after this refactor and associated pub change)
- external repo IDs

I suppose it's possible UD was updated to accept those types of repo IDs and I don't know about it.  On the other hand it's possible this code doesn't make sense.  Could anyone explain it?

rohanpm:
Counterpart of Pub change https://code.engineering.redhat.com/gerrit/c/rcm-pub/+/247503.

midnightercz (approved):

pubtools/_quay/push_docker.py
rohanpm: Note, I have https://code.engineering.redhat.com/gerrit/c/rcm-pub/+/248049 also (but still waiting for CI)

PR #34, opened 2021-06-15
Read extra tokens for osbs organizations

emilyzheng:
Use different robot accounts and OAuth tokens for source images and destination images.


pubtools/_quay/container_image_pusher.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, push_items, target_settings):
 
         self.quay_host = self.target_settings.get("quay_host", "quay.io").rstrip("/")
         self._quay_client = None
+        self._quay_client_osbs = None
```
querti, line 44: I would prefer if the names were something like "source_quay_client" and "dest_quay_client". It would be more self-explanatory that way

pubtools/_quay/signature_handler.py
```diff
@@ -548,7 +548,7 @@ def construct_index_image_claim_messages(self, index_image, tag, signing_keys):
         internal_repo_schema = self.target_settings["quay_namespace"] + "/{internal_repo}"
 
         # Get digests of all archs this index image was build for
-        manifest_list = self.quay_client.get_manifest(index_image, manifest_list=True)
```
querti, line 551: Shouldn't ManifestListMerger be modified as well?

querti (changes requested):
One more change is probably necessary - skopeo copy logic. Since copying is done between two different organizations, two distinct skopeo logins will be necessary. As a consequence, tag_images entrypoint has to be modified as well. I feel like creating a new Jira just for this change is appropriate. What do you think?

emilyzheng:
> One more change is probably necessary - skopeo copy logic. Since copying is done between two different organizations, two distinct skopeo logins will be necessary. As a consequence, tag_images entrypoint has to be modified as well. I feel like creating a new Jira just for this change is appropriate. What do you think?

skopeo copy from registry-proxy doesn't need credential of it, so only login of dest image is required.

querti:
> > One more change is probably necessary - skopeo copy logic. Since copying is done between two different organizations, two distinct skopeo logins will be necessary. As a consequence, tag_images entrypoint has to be modified as well. I feel like creating a new Jira just for this change is appropriate. What do you think?
> 
> skopeo copy from registry-proxy doesn't need credential of it, so only login of dest image is required.

Hmm so do we even need two robot accounts at all? Or do some operations need authorization on registry proxy?

emilyzheng:
> Hmm so do we even need two robot accounts at all? Or do some operations need authorization on registry proxy?

get_manifest from registry proxy needs robot account, but skopeo copy not (like pulling image from registry proxy).

pubtools/_quay/container_image_pusher.py
emilyzheng: Updated.

querti (approved):

midnightercz (approved):

PR #38, opened 2021-06-24
Change "repo" parameter of claim messages to have external represention

querti:
This change should not affect functionality, it will just ensure
that metadata construction of claim messages is better aligned with
the current docker-pulp implementation.


midnightercz:
hmm, pub docker-pulp implementation use pulp repositories for "repo" attribute, so it is basically internal format

querti:
> hmm, pub docker-pulp implementation use pulp repositories for "repo" attribute, so it is basically internal format

'repo' and 'image_name' fields have the same value in the pulp2 messages, while they have a different value here. We've felt that it would be safer to keep the expected behavior

emilyzheng (approved):

midnightercz (approved):

PR #44, opened 2021-07-07
Set signing key for operator push items

emilyzheng:
Operator push items from an advisory push don't have claims_signing_key.
Set it with signing key of docker push items.

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -147,7 +147,7 @@ class PushDocker:
         return docker_push_items
 
     @log_step("Get operator push items")
-    def get_operator_push_items(self):
+    def get_operator_push_items(self, docker_push_items):
         """
         Filter out push items to only include operator ones.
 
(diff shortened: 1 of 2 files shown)


pubtools/_quay/push_docker.py
```diff
@@ -178,6 +178,8 @@ def get_operator_push_items(self):
             if not item.metadata.get("com.redhat.openshift.versions"):
                 msg = "Push item {0} doesn't specify 'com.redhat.openshift.versions'".format(item)
                 raise BadPushItem(msg)
+            if not item.claims_signing_key and docker_push_items:
```
querti, line 181: What happens when there are no docker push items?

emilyzheng: If a brew build having only operator item exists, IIB request before signing would fail because bundle to be added is not available.

querti (approved):

pubtools/_quay/push_docker.py
```diff
@@ -178,6 +178,8 @@ def get_operator_push_items(self):
...
+                item.claims_signing_key = docker_push_items[0].claims_signing_key
```
midnightercz, line 182: I think it would rather make sense to do this change here: pubd/pubd/push/adapter.py

emilyzheng:
The change was done in Pub.

PR #45, opened 2021-07-08
Allow pushing to not existing repo

emilyzheng:
Response code of get_manifest request is 401 when a repo doesn't exist.

### DIFF
--- a/pubtools/_quay/container_image_pusher.py
+++ b/pubtools/_quay/container_image_pusher.py
@@ -217,7 +217,7 @@ class ContainerImagePusher:
                         merge_mls_dest_refs.append(dest_ref)
                 except requests.exceptions.HTTPError as e:
                     # Option 3: Destination tag doesn't exist, no ML merging
-                    if e.response.status_code == 404:
+                    if e.response.status_code == 404 or e.response.status_code == 401:
                         simple_dest_refs.append(dest_ref)
                     else:
                         raise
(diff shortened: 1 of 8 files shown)


pubtools/_quay/container_image_pusher.py
```diff
@@ -217,7 +217,7 @@ def copy_multiarch_push_item(self, push_item, source_ml):
                         merge_mls_dest_refs.append(dest_ref)
                 except requests.exceptions.HTTPError as e:
                     # Option 3: Destination tag doesn't exist, no ML merging
-                    if e.response.status_code == 404:
+                    if e.response.status_code == 404 or e.response.status_code == 401:
```
querti, line 220: I think the code relies on 404 in multiple places in the code. Shouldn't we change it everywhere?

querti (changes requested):

pubtools/_quay/container_image_pusher.py
emilyzheng: You are right. I will update other places. Do you think 404 should be kept or removed?

querti: I think it wouldn't hurt to keep it. It only gives 401 when a robot account is used, I think

emilyzheng: Updated.

querti (approved):

midnightercz (approved):

PR #46, opened 2021-07-08
Add entrypoint to remove all references of an image

querti:

PR #47, opened 2021-07-14
Revert "Add entrypoint to remove all references of an image"

querti:
Reverts release-engineering/pubtools-quay#46


querti:
Reverting the new entrypoint since it will not be used.

PR #53, opened 2021-07-29
Add option to execute commands inside a container

querti:
Also, update the documentation reflecting the new changes, define
a Dockerfile of image capable of running all necessary skopeo
operations, and start using executors as context managers
(to perform self-cleanup).


pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,163 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+                Path to an image which will be used for performing skopeo operations. Must be
+                downloadable by Docker.
+            base_url (str):
+                Base URL of the Docker client.
+            timeout (int):
+                Default timeout for API calls, in seconds.
+            verify_tls (bool):
+                Whether to use TLS verification.
+            cert_path (str|None):
+                Custom path to TLS cettificates.
```
neithere, line 288: `s/tt/rt/`

also it could be useful to mention that `~/.docker/` is used by default. Maybe even make it the default arg value instead of `None` and simply `os.path.expanduser(cert_path)`?

pubtools/_quay/tag_docker.py
```diff
@@ -734,34 +722,46 @@ def run(self):
...
+            self.target_settings["skopeo_image"],
+            self.target_settings.get("docker_host") or "unix://var/run/docker.sock",
+            self.target_settings.get("docker_timeout"),
+            self.target_settings.get("docker_tls_verify") or False,
+            self.target_settings.get("docker_cert_path") or None,
+        ) as executor:
+            executor.skopeo_login(
+                self.target_settings["dest_quay_user"], self.target_settings["dest_quay_password"]
+            )
+            for item in self.push_items:
```
neithere, line 735: Is there a chance for the `run()` method to include more conditions over time? In that case it may eventually become hard to understand even despite the pretty good documentation, so it would make sense to extract the loop body to separate method(s). Otherwise it's fine.

pubtools/_quay/tag_images.py
```diff
@@ -199,26 +251,42 @@ def tag_images(
...
+            ContainerExecutor,
+            container_image,
+            docker_url,
+            docker_timeout,
+            docker_verify_tls,
+            docker_cert_path,
+        )
     else:
-        executor = LocalExecutor()
+        executor_class = functools.partial(LocalExecutor)
```
neithere, line 284: Isn't this a longer way to say `executor_class = LocalExecutor`? :)

pubtools/_quay/tag_images.py
```diff
@@ -199,26 +251,42 @@ def tag_images(
...
 
-    executor.skopeo_login(quay_user, quay_password)
-    executor.tag_images(source_ref, dest_refs, all_arch)
+    with executor_class() as executor:
+        print(executor)
```
neithere, line 287: Debug print?

docker/skopeo_runner/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM fedora:33
```
rohanpm, line 1: Is this the image you intend to use in production?  Can you share some info on how it'll be built?  It is best to have some kind of plan how to ensure this will be regularly rebuilt, and updated when F33 isn't supported any more.

In another repo I've had success with approach:

- use a specific digest in FROM
- activate renovatebot
- it will automatically bump the digest when a new version is available, e.g. https://github.com/rohanpm/ssh-prometheus-exporter/pull/9
- github updates trigger quay rebuild at https://quay.io/repository/rmcgover/ssh-prometheus-exporter

That way the repo keeps track of a known good image, and there's some automated prompt to get it regularly rebuilt.

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,163 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+                kwargs["tls"] = docker.tls.TLSConfig(
+                    client_cert=(
+                        os.path.join(cert_path, "cert.pem"),
+                        os.path.join(cert_path, "key.pem"),
+                    ),
+                    verify=os.path.join(cert_path, "ca.pem"),
+                )
+
+        self.client = docker.APIClient(**kwargs)
+        self.container = self.client.create_container(self.image, detach=True, tty=True)
```
rohanpm, line 309: What kind of values are you intending to provide as the image ?

If you want the possibility of images being updated after first use, should there be some 'pull' operation here?  On the other hand, the convention could be that you just have to update image in target settings if you ever want it to be updated.  But I have some doubts whether anyone would regularly do that.

docker/skopeo_runner/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+# TODO: should the image contain any special labels?
```
zanssa, line 3: It's a metadata about the image. So, we could start with adding maintainer & description labels. 

LABEL Maintainer=" " \
           Description=" "

docker/skopeo_runner/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+WORKDIR /src
+
+RUN dnf install -y skopeo
```
zanssa, line 7: in docker images, I prefer to do the installation & clean up in one layer.

zanssa, line 7:
```suggestion
RUN dnf install -y skopeo && dnf -y clean all
```

docker/skopeo_runner/Dockerfile
zanssa: you could also use github action to start building the image & pushing it to quay whenever there is a change in your dockerfile. here is an example: https://github.com/zanssa/iib-deployment/blob/main/.github/workflows/quay.yml

twaugh:
Naively asking: is there a reason `docker` is being used instead of `podman`?

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,163 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+            LOG.info("Already logged in to Quay.io")
+            return
+
+        if not username or not password:
+            raise ValueError(
+                "Skopeo login credentials are not present. Quay user and token must be provided."
+            )
+        LOG.info("Logging in to Quay with provided credentials")
+
+        password_file = "skopeo_password.txt"
```
midnightercz, line 405: did you consider using docker ENV ?

midnightercz (changes requested):

pubtools/_quay/command_executor.py
querti: Fixed typo and added information about default location. This part is mostly copy-pasted from pub so I'd prefer to leave it as is, if that's okay.

pubtools/_quay/tag_docker.py
querti: I don't expect this logic to become more complex over time, but you never know. I'd say let's leave the refactoring out of this PR and do it once it becomes necessary. What do you think?

pubtools/_quay/tag_images.py
querti: :D . But I'd prefer to keep it this way because maybe LocalExecutor will accept some constructor arguments in the future. Also, I feel the code is a bit less convoluted when all executors are created the same way. Is it okay?

pubtools/_quay/tag_images.py
querti: Yep, removed.

docker/skopeo_runner/Dockerfile
querti: I added renovatebot settings and Github action for building it (just as a test). I'll add a Quay trigger after this PR is merged.

pubtools/_quay/command_executor.py
querti: I added a pull operation. Testing indicates that it's not a very expensive operation if the image is already downloaded.

docker/skopeo_runner/Dockerfile
querti: Added

querti: Added

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,163 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+                    client_cert=(
+                        os.path.join(cert_path, "cert.pem"),
+                        os.path.join(cert_path, "key.pem"),
+                    ),
+                    verify=os.path.join(cert_path, "ca.pem"),
+                )
+
+        self.client = docker.APIClient(**kwargs)
+        self.container = self.client.create_container(self.image, detach=True, tty=True)
+        self.client.start(self.container["Id"])
```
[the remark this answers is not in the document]
querti: But will it be triggered if the executor is not used as context manager?

pubtools/_quay/command_executor.py
querti: I suppose it would be a viable solution as well, though I would prefer to keep the current one, simply because it's already implemented and tested. Is that ok?

pubtools/_quay/command_executor.py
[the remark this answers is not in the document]
midnightercz: It won't but you have stuff in __exit__ anyway, so you need to use it as context manager

pubtools/_quay/command_executor.py
midnightercz: sure, I was just wondering if was any problem with ENV or anything which prevented you from using it

midnightercz (approved):

querti:
> Naively asking: is there a reason `docker` is being used instead of `podman`?

First reason is that pub-docker-daemon uses docker, not podman. Second reason is that Docker SDK for Python is used (https://docker-py.readthedocs.io/en/stable/), and I'm not sure if it's actually compatible with podman. It very well might be though.

pubtools/_quay/command_executor.py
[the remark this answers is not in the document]
querti: Well, exit only performs a cleanup, so it can be used as a regular class if the user doesn't mind zombie containers being left behind.

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,165 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+            stdin (None|str):
+                This parameter exists only for compatibility with parent class. Sending input
+                to containers is not supported.
+
+        Returns (str, str):
+            Tuple of stdout and stderr generated by the command.
+        """
+        err_msg = err_msg or "An error has occured when executing a command."
+        cmd_exec = self.client.exec_create(self.container["Id"], cmd)
+        stdout, stderr = self.client.exec_start(cmd_exec["Id"], demux=True)
```
emilyzheng, line 338: exec_api installed by python-docker-py-1.7.2-1.el6eng.noarch doesn't have argument 'demux'.

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,165 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+            if os.path.isdir(cert_path):
+                kwargs["tls"] = docker.tls.TLSConfig(
+                    client_cert=(
+                        os.path.join(cert_path, "cert.pem"),
+                        os.path.join(cert_path, "key.pem"),
+                    ),
+                    verify=os.path.join(cert_path, "ca.pem"),
+                )
+
+        self.client = docker.APIClient(**kwargs)
```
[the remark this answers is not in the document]
querti: Added the correct import based on the version of docker.

pubtools/_quay/command_executor.py
```diff
@@ -245,3 +258,165 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+            timeout (int):
+                Default timeout for API calls, in seconds.
+            verify_tls (bool):
+                Whether to use TLS verification.
+            cert_path (str|None):
+                Custom path to TLS certificates. If not specified, '~/.docker' is used.
+        """
+        self.image = image
+
+        kwargs = {}
```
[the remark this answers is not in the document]
querti: I checked target settings and it seems that specific version is not actually set, so "auto" is always used. I added an argument to always use "auto".

pubtools/_quay/command_executor.py
querti: Good point, I removed the parameter. The unfortunate side effect is that now we can't differentiate stdout and stderr. To maintain consistency with other classes, I duplicated the output to be returned as tuple. I think this change shouldn't affect any of the current use-cases.

pubtools/_quay/command_executor.py
```diff
@@ -5,16 +5,24 @@
...
 import docker
 import paramiko
+from six.moves import shlex_quote
 
 LOG = logging.getLogger("pubtools.quay")
 
 
+# Docker client is called differently based on the version used. Unify the calls.
+# Only use old call if version 1.X.X is used
+class APIClient(docker.APIClient if int(docker.__version__[0]) > 1 else docker.client.Client):
```
neithere, line 20: Not sure if inheritance is better than plain assignment here.

pubtools/_quay/command_executor.py
```diff
@@ -319,6 +328,11 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
         """
         Run a command locally.
 
+        NOTE: Older versions of Docker API don't suppoer demuxing of stdout and stderr.
```
neithere, line 331: s/suppoer/support/

pubtools/_quay/command_executor.py
```diff
@@ -335,20 +349,18 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
         """
         err_msg = err_msg or "An error has occured when executing a command."
         cmd_exec = self.client.exec_create(self.container["Id"], cmd)
-        stdout, stderr = self.client.exec_start(cmd_exec["Id"], demux=True)
+        # Unfortunately, older versions of Docker API don't support demuxing stdout and stderr
```
neithere, line 352: Shouldn't the result depend on the version?

neithere (approved):

pubtools/_quay/command_executor.py
querti: fixed

pubtools/_quay/command_executor.py
querti: The main issue is that older versions of the Python dependency didn't even support this option.

PR #54, opened 2021-08-03
Use SSL certificates for Pyxis authentication

querti:
For entrypoints, change the user arguments to include certificates.
For code invoked by tasks, use hooks to get file location from pub.


midnightercz (approved):

pubtools/_quay/hooks.py
```diff
@@ -47,4 +47,13 @@ def quay_images_untagged(untag_refs, lost_refs):
     """
 
 
+@hookspec
+def get_cert_key_paths_plugin(server_url):
```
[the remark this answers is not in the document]
querti: In pub. The function is 'get_cert_key_paths', which is used in multiple places in the code. I thought using hooks would be a good way to prevent code duplication

pubtools/_quay/iib_operations.py
```diff
@@ -137,12 +139,14 @@ def task_iib_add_bundles(
         quay_user=target_settings["dest_quay_user"],
         quay_password=target_settings["dest_quay_password"],
     )
+    cert, key = pm.hook.get_cert_key_paths_plugin(server_url=target_settings["pyxis_server"])
```
midnightercz, line 142: if pubtools-quay is running without pub, this should return None, if it's not registered right? Or am I missing something? In that case we should add fallback to target_settings

midnightercz (changes requested):

pubtools/_quay/iib_operations.py
rohanpm: A couple of things here:

1) Not sure if this actually works.  Was it realistically tested? The hook call doesn't directly return the result, it returns a list of results, since there can be multiple hookimpl attached.  See https://pluggy.readthedocs.io/en/latest/#collecting-results and also see "firstresult" if you want to use that to handle the "more than one hookimpl" scenario.

See also my later comment in a test where pm was mocked.

2) I agree that there should be some fallback or way of making this work outside of Pub.

Actually this usage of hooks doesn't match what I had in mind when introducing the system originally, and I'm having trouble forming an opinion whether this is a good idea.  But I'm confident at least that it shouldn't be the *only* way this code can be reasonably used because that goes against some of the goals of splitting our projects up.

It's intended that each of the projects like pubtools-quay etc should be usable standalone, not only inside of Pub. 
 If we set things up like "this code *requires* a hook to return a value, and the only implementer lives in Pub" then that's pushing us in the opposite direction - we're creating more reasons that this code can't possibly work unless you run it in Pub.

I don't know exactly what the default behavior should be, but it should be something not overly cumbersome, e.g. if I've got a shell open and I want to run "pubtools-quay-(something)" and I want to make it use a certain cert and key, there should be a way I can do that.

tests/test_iib_operations.py
```diff
@@ -33,6 +33,7 @@ def test_verify_target_settings_overwrite_index_mismatch(target_settings):
         iib_operations.verify_target_settings(target_settings)
 
 
+@mock.patch("pubtools._quay.iib_operations.pm")
```
rohanpm, line 36: Why mock this?

It's just an object with a well-defined API whose state you can modify locally, with no complicated setup needed, no dependency on remote systems etc.  When you could easily test with the real object, I don't get why you want to replace it with a mock - by doing that, you open the possibility of configuring the mock incorrectly.  It seems that this did in fact happen (the return_value in this test is impossible, see earlier comment about it returning a list) which allowed the change overall to be broken, yet passing CI.

Possibly there could be a mix of mocking and not mocking across different tests, but I think you should have at least one test which actually uses it in a realistic manner.  This doesn't seem difficult, isn't it approximately this:

```python
# helper class somewhere
class FakePaths:
  @hookimpl
  def get_cert_key_paths_plugin(self, server_url):
    return ("/path/to/file.crt", "/path/to/file.key")

@pytest.fixture
def fake_paths():
    # Installs a hookimpl for cert/key for the duration of a test.
    fake_paths = FakePaths()
    pm.register(fake_paths)
    yield
    pm.unregister(fake_paths)

# now any test which accepts 'fake_paths' argument will have that hookimpl
# registered during the test
```

Then your tests would be exercising the real hook mechanism, giving you confidence that it works.
Rather than exercising a mock object that you assert should match the real hook mechanism, but in reality might not.

rohanpm (changes requested):

pubtools/_quay/hooks.py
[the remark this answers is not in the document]
querti: I decided to go the other direction and remove and reference to Pyxis. The function in pub is generic and should work for any service, so I decided that pubtools-quay's documentation should reflect that, even though the function only used for Pyxis right now.

pubtools/_quay/iib_operations.py
querti: You're right, during testing I mainly focused on the implementation being properly registered from pub, I didn't pay much attention to the returned value. I decided to use the first returned value because I wanted to make it "best effort"  even though it's debatable whether more than one implementation should even be allowed. If there are no returned values, I decided to use target settings values. I know you're not a big fan of us using target settings, its replacement could be a part of some future improvement.

pubtools/_quay/hooks.py
```diff
@@ -47,4 +47,13 @@ def quay_images_untagged(untag_refs, lost_refs):
...
+    """Get location of SSL certificates. Method expected to be implemented by pub.
+
+    :param server_url: Pyxis hostname
```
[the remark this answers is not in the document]
querti: url

pubtools/_quay/hooks.py
```diff
@@ -47,4 +47,13 @@ def quay_images_untagged(untag_refs, lost_refs):
     """
 
 
+@hookspec
+def get_cert_key_paths_plugin(server_url):
+    """Get location of SSL certificates. Method expected to be implemented by pub.
```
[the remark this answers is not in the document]
querti: I documented the decided behavior. Are you okay with this approach?

tests/test_iib_operations.py
querti: Modified tests to use real hooks.

pubtools/_quay/hooks.py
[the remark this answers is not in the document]
rohanpm: If you want to position this as a generic hook, can you please add it to `pubtools` instead?

It doesn't seem like a good idea to have a hook which is intended to be shared (i.e. hook possibly invoked from outside of pubtools-quay), but the hookspec lives in pubtools-quay project.  I expect it risks some problems such as projects working differently depending on whether pubtools-quay was installed & imported.  At the very least it gives pubtools-quay the ability to break all the projects using it by making backwards-incompatible changes to the hookspec.

I'm warming up to the idea of having this as a generic hook in pubtools... if you do want to go through with that, may I suggest some other minor changes:

- should it have 'firstresult' in the hookspec?
- there are many possible formats of cert/key, so it'd probably be worthwhile for the doc to clarify that we're expecting PEM-formatted files ("BEGIN CERTIFICATE" and "BEGIN RSA PRIVATE KEY") and maybe also that it's permitted for both paths to be the same.
- the fallback strategy part should probably be changed to not mention target settings.  If it's generic, it doesn't make sense to say "...and here's what pubtools-quay does if this hook isn't implemented".  I think the doc string should be written with the primary target audience being whoever wants to invoke the hook, so from their point of view, the relevant information is something like: "Callers of this hook should be prepared to receive no result, and should implement a reasonable fallback strategy in that case."

pubtools/_quay/utils/misc.py
```diff
@@ -284,3 +285,35 @@ def fn_wrapper(*args, **kwargs):
...
+    was registered last). Otherwise, raise an error.
+
+    Args:
+        target_settings (dict):
+            Dictionary containing various task-related settings.
+    Returns ((str, str)):
+        Paths to Pyxis SSL certificate and key.
+    """
+    result = pm.hook.get_cert_key_paths_plugin(server_url=target_settings["pyxis_server"])
+    if result:
```
rohanpm, line 307: This will use the first result, but I think you should instead use the first non-None result.  Reason for this is that it allows hooks to indicate "sorry, I don't know what to do here" - e.g. a hook which knows how to get cert/key for some services and not others - and then let another hook take over.

This brings me back to 'firstresult' again because https://pluggy.readthedocs.io/en/latest/#collecting-results says that's exactly what happens if you use firstresult on the hookspec.

rohanpm (changes requested):

pubtools/_quay/hooks.py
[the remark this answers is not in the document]
querti: Created a PR for pubtools. I think this PR is blocked until a new pubtools release, not sure how to test the hook otherwise.

pubtools/_quay/clear_repo.py
```diff
@@ -47,14 +47,14 @@
...
-        "help": "Pyxis kerberos principal in form: name@REALM",
+    ("--pyxis-ssl-crtfile",): {
+        "help": "Path to .crt file for the SSL authentication",
         "required": True,
         "type": str,
     },
-    ("--pyxis-krb-ktfile",): {
-        "help": "Pyxis Kerberos client keytab. Optional. Used for login if TGT is not available.",
-        "required": False,
+    ("--pyxis-ssl-keyfile",): {
```
rohanpm, line 55: Realized while reading the corresponding Pub change:

Why does this command take the cert/key as arguments rather than using the hook?  Is it intentional that some of the commands in pubtools-quay will use the hook to get the cert/key paths, and other commands will instead use command-line arguments?  Why is that?

querti: Well, I recall that at one point you mentioned you're disappointed that we use target settings when invoking pubtools quay. So, for the remaining entrypoints I tried to remove dependency on pub (it was also easier, I didn't required instance of hub XMLRPC). Thus, clear and remove repo are implemented as CLI applications, and pub constructs arguments when invoking them. This is unlike other entrypoints invoked by pub, which require some not so CLI friendly parameters (target settings, and instance of hub). When using arguments, pub could resolve the paths and specify them when invoking the new entrypoints. One day we could perhaps unify these approaches.

rohanpm: > When using arguments, pub could resolve the paths and specify them when invoking the new entrypoints.

Yeah, it could, but the question is whether it *should* now that the hook exists and is used elsewhere.

Ultimately I think this is a matter for the pubtools-quay maintainers and I don't have a strong opinion either way, but I do think this is being set up less maintainable than it could be, unnecessarily.

You're setting things up so there are two places where Pub is being coupled with this project - it has to *both* implement the hook *and* pass certain arguments into some commands in order for cert/key handling to work properly.  That's two places which need to be maintained or extended/changed if requirements change.  You could easily make it just one place (Pub only has to implement the hook).

querti: I thought you wanted to go the other direction. The hook doesn't exactly solve anything since it requires a fallback mechanism (target settings at the moment). Plus, it implies strong coupling with pub. I thought that once we have establish a pub-independent way of passing many parameters we will use that instead of hooks. Since you don't have a strong opinion either way, I won't consider this a blocker.

midnightercz (approved):

emilyzheng (approved):

PR #56, opened 2021-08-05
Add pagination support to get all tags Docker HTTP API call

querti:


pubtools/_quay/quay_client.py
```diff
@@ -145,11 +145,19 @@ def get_repository_tags(self, repository, raw=False):
         """
         endpoint = "{0}/tags/list".format(repository)
         response = self._request_quay("GET", endpoint)
+        tags = response.json()
+
+        while "Link" in response.headers:
+            # next page response has format '</v2/....>; rel="next"'
+            # so, split by ; -> remove <> from url -> remove v2 prefix from url
```
neithere, line 152: Wouldn't it be safer to use a full regex like `</v2/(.+?)>; rel="next"` than to rely on positions?
That said, we probably shouldn't expect anything else here.

neithere (approved):
LGTM.

pubtools/_quay/quay_client.py
zanssa: Agreed. You could import `re` library and do like Andy suggested or something like this to extract the next page: 

``` 
test = re.findall('</v2/(.*)>', Link)
print(test)
['namespace/image/tags/list/next-page-2']
```

PR #58, opened 2021-08-10
Capture IIB operation exception

emilyzheng:
sys.exit() is called when IIB operation fails. Capture the
exception raised by it so that rollback is not skipped for
IIB failures.

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -550,7 +550,7 @@ class PushDocker:
                 operator_signature_handler.sign_operator_images(iib_results)
                 # Push index images to Quay
                 operator_pusher.push_index_images(iib_results)
-        except Exception:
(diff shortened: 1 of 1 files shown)


querti (approved):
Is it raised by pubtools-iib? It seems quite odd to raise an error like this. @midnightercz Is there a particular reason for it? If not, I'd suggest we change it in pubtools-iib.

emilyzheng:
> Is it raised by pubtools-iib? It seems quite odd to raise an error like this. @midnightercz Is there a particular reason for it? If not, I'd suggest we change it in pubtools-iib.

It's from https://github.com/release-engineering/pubtools-iib/blob/master/pubtools/iib/iib_ops.py#L262.

midnightercz:
@querti I think it was configured like that for pub

midnightercz (approved):

PR #69, opened 2021-08-19
Run rollback only when all index image builds fail

midnightercz:
If we rollback when one of the index image builds is successful,
all bundles referenced by it are not available.\

relevant JIRA: CLOUDDST-3246


querti (changes requested):
It seems like you removed rollback completely and only allowed it for one very specific scenario, when all index image builds fail. Are you sure that's what you wanted to do?

midnightercz:
> It seems like you removed rollback completely and only allowed it for one very specific scenario, when all index image builds fail. Are you sure that's what you wanted to do?
Yeah, I think this is what we want. 1. It doesn't make sense to do rollback when there are no operators. 2. According to jira description, we want to rollback only when everything fails, otherwise some bundle referenced by index image which is already pushed won't be available.
Another option is to also rollback index images, but I see that quite dangerous
cc @yashvardhannanavati

querti (approved):

pubtools/_quay/push_docker.py
```diff
@@ -538,49 +539,53 @@ def run(self):
...
+            # Push index images to Quay
+            operator_pusher.push_index_images(successful_results, index_stamp)
+            # Rollback only when all index image builds fails
+            if set([x["iib_result"] for x in iib_results.values()]) == set([False]):
+                LOG.error("Push of all index images failed, running rollback.")
+                self.rollback(backup_tags, rollback_tags)
+                sys.exit(1)
+            if successful_results != iib_results:
+                LOG.error("Push of some index images failed")
+                sys.exit(1)
```
lipoja, line 576: I am just wondering if it is safe to use `sys.exit()` instead of `return` or `raise`?

pubtools/_quay/push_docker.py
```diff
@@ -538,49 +539,53 @@ def run(self):
...
+            operator_pusher = OperatorPusher(operator_push_items, self.target_settings)
+            existing_index_images = operator_pusher.get_existing_index_images(self.dest_quay_client)
+            iib_results = operator_pusher.build_index_images()
+            # Sign operator images
+            successful_results = dict(
+                [(key, val) for key, val in iib_results.items() if val["iib_result"]]
             )
+            operator_signature_handler.sign_operator_images(successful_results, index_stamp)
+            # Push index images to Quay
+            operator_pusher.push_index_images(successful_results, index_stamp)
```
emilyzheng, line 568: push_index_images has:

```
        for version in self.version_items_mapping:
            build_details = iib_results[version]["iib_result"]
```
It complains when a version is not in results.

pubtools/_quay/push_docker.py
querti: Is there a reason for using sys.exit?

midnightercz: we catch for SystemExit in pub

pubtools/_quay/push_docker.py
midnightercz: fixed

querti (changes requested):

pubtools/_quay/push_docker.py
querti: Could you elaborate on that?

midnightercz: in EntryPointBase there's code in run method:
except SystemExit as e:
...

pubtools/_quay/push_docker.py
```diff
@@ -537,50 +538,55 @@ def run(self):
...
             )
+            operator_signature_handler.sign_operator_images(successful_iib_results, index_stamp)
+            # Push index images to Quay
+            operator_pusher.push_index_images(successful_iib_results, index_stamp)
+            # Rollback only when all index image builds fails
+            if set([x["iib_result"] for x in iib_results.values()]) == set([False]):
+                LOG.error("Push of all index images failed, running rollback.")
+                self.rollback(backup_tags, rollback_tags)
+                sys.exit(1)
+            if successful_iib_results != iib_results:
```
emilyzheng, line 575: Should remove_old_signatures be moved before this? Or it will be skipped in this case.

pubtools/_quay/push_docker.py
querti: But is there a particular reason for this solution? What would happen if you just raised a normal error?

pubtools/_quay/push_docker.py
midnightercz: Yes, I guess it can exit after old signatures removal.

PR #79, opened 2021-08-27
Tag and sign timestamp for pushed index images

midnightercz:
In order to avoid garbage collection of old tags for index images
which are regularly overwritten by every push, timestmap tags are
used to create unique tag for every index image pushed.
Timestamp tags are applied per index image version in format
version-stamp, for example: v4.9-1630072512

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -5,7 +5,7 @@ from .exceptions import InvalidTargetSettings
 from .operator_pusher import OperatorPusher
 from .signature_handler import OperatorSignatureHandler
 from .signature_remover import SignatureRemover
-from .utils.misc import get_internal_container_repo_name, get_pyxis_ssl_paths
+from .utils.misc import get_internal_container_repo_name, get_pyxis_ssl_paths, timestamp
 
 LOG = logging.getLogger("pubtools.quay")
 
@@ -127,8 +127,11 @@ def task_iib_add_bundles(
     )
 
     # Sign image
+    index_stamp = timestamp()
     sig_handler = OperatorSignatureHandler(hub, task_id, target_settings, target_name)
-    claim_messages = sig_handler.sign_task_index_image(signing_keys, intermediate_index_image, tag)
+    claim_messages = sig_handler.sign_task_index_image(
+        signing_keys, intermediate_index_image, tag, "%s-%s" % (tag, index_stamp)
+    )
 
     sig_remover = SignatureRemover(
         quay_api_token=target_settings["dest_quay_api_token"],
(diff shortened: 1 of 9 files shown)


pubtools/_quay/iib_operations.py
```diff
@@ -127,8 +127,11 @@ def task_iib_add_bundles(
     )
 
     # Sign image
+    index_stamp = timestamp()
```
querti, line 130: I wonder, is it possible that two images may end up with the same tag?

pubtools/_quay/operator_pusher.py
```diff
@@ -414,7 +414,7 @@ def build_index_images(self):
         return iib_results
 
     @log_step("Push index images to Quay")
-    def push_index_images(self, iib_results):
+    def push_index_images(self, iib_results, timestamp=None):
```
querti, line 417: Missing docstring

querti: Also, I feel like this name is too tied to the current implementation. I would maybe rename it to "extra_tag" or "redundant_tag" or something. I think calling it timestamp conveys information that's irrelevant in this context.

pubtools/_quay/signature_handler.py
```diff
@@ -503,14 +503,16 @@ def sign_container_images(self, push_items):
...
-    def construct_index_image_claim_messages(self, index_image, tag, signing_keys):
+    def construct_index_image_claim_messages(self, index_image, tag, stamp_tag, signing_keys):
         """
         Construct signature claim messages for RADAS for the specified index image.
 
         index_image (str):
             Reference to a new index image constructed by IIB.
         tag (str):
             Tag of the newly built index image.
+        tag (str):
```
querti, line 514: Wrong docstring

pubtools/_quay/signature_handler.py
```diff
@@ -503,14 +503,16 @@ def sign_container_images(self, push_items):
 class OperatorSignatureHandler(SignatureHandler):
     """Class for handling the signing of index images."""
 
-    def construct_index_image_claim_messages(self, index_image, tag, signing_keys):
+    def construct_index_image_claim_messages(self, index_image, tag, stamp_tag, signing_keys):
```
querti, line 506: I think it would be better to allow list to be passed to "tags" parameter, instead of having two distinct parameters for tags. That way, when reading the method, the user doesn't have to wonder what the heck is a stamp_tag.

pubtools/_quay/signature_handler.py
```diff
@@ -536,21 +538,24 @@ def construct_index_image_claim_messages(self, index_image, tag, signing_keys):
...
+                            image_name=repo,
+                            task_id=self.task_id,
+                        )
+                        claim_messages.append(claim_message)
 
         return claim_messages
 
     @log_step("Sign operator images")
-    def sign_operator_images(self, iib_results):
+    def sign_operator_images(self, iib_results, index_stamp):
```
querti, line 558: missing docstring

querti: Also, I would probably rename this variable. The fact that it's a timestamp is not relevant here, only that it's a redundant tag

querti (changes requested):

pubtools/_quay/iib_operations.py
querti: What about two parallel pushes?

midnightercz: I think there's almost 0 chance it can happen.

emilyzheng (approved):

querti (approved):
Has it been decided whether we'll go with this approach?

PR #81, opened 2021-09-02
Get container push items by unique pull_url

emilyzheng:
Container push items added from a multi-arch brew build have the
same push source and targets. Include only one of them as an item
for operations like signing, pushing, etc. so no duplicate content
are dealt with.

Refers to CLOUDDST-7152.

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -126,10 +126,13 @@ class PushDocker:
 
         Also, Check the validity of these items and raise an exception in case of incorrect data.
 
+        For items having the same pull_url, only one of them is returned to avoid duplication.
+
         Returns ([ContainerPushItem]):
             Docker push items.
         """
         docker_push_items = []
+        url_items = {}
         for item in self.push_items:
             if item.file_type != "docker":
                 LOG.warning("Push item {0} doesn't have 'docker' type, skipping.".format(item))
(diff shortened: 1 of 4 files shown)


pubtools/_quay/push_docker.py
```diff
@@ -139,7 +142,13 @@ def get_docker_push_items(self):
...
                 raise BadPushItem("Push item {0} doesn't contain pull data.".format(item))
             LOG.info("Docker push item found: {0}".format(item))
-            docker_push_items.append(item)
+            pull_url = item.metadata["pull_url"]
+            if pull_url in url_items:
+                url_items[pull_url].append(item)
+            else:
+                url_items[pull_url] = [item]
+        for _, items in sorted(url_items.items()):
+            docker_push_items.append(items[0])
```
querti, line 151: Can't push items have other differences than pull url? For example ["tags"] field tells us where to push the image. We would this information if we filtered items only by unique pull url.

Or am I misunderstanding something?

querti (changes requested):

pubtools/_quay/push_docker.py
emilyzheng: They have difference on some other metadata, e.g. dev_tags, arch, but not on tags.

querti: @midnightercz you probably understand best how sources generate push items, is this safe?

midnightercz: Container push items are populated here: 
https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/backend/koji_source.py#L403
Which is also where this filter should be implemented. Functionality is split into source/target so having this implemented here mixes those to together.

midnightercz (changes requested):

pubtools/_quay/push_docker.py
midnightercz: But besides that, change is ok. Destination tags are set in target, so if push items share source url, it's ok to filter duplicates

emilyzheng: I doubt it's wanted filter for Pulp targets. Pushes to pulp pull images from url (digest of each arch) in image archive , not the url (manifest list) in build information.

midnightercz (approved):

querti (approved):

PR #95, opened 2021-11-05
Make IIB tag new index images with unique tag and use it to get manifest

querti:
A while ago it was discovered that using floating tag of index images
to get manifest list may return the incorrect result due to a race
condition (index image was overwritten by a parallel push). The solution
was to use intermediate index image specified by digest. However, as
this index image may be overwritten as well, the new solution is to
make IIB tag the new index image with a unique tag and use it to
get the manifest list for signing. Tag uniqueness is ensured by
combining the index image floating tag (OCP version) and pub task
ID. This change was applied to all IIB calls done by pubtools-quay,
so standard push-docker workflow and separate IIB-only workflows.

Few minor unrelated changes were done as well, namely cleanup of
unused variables, docstring fixes, and change of "failed" variable
in push_docker.py to bool instead of integer because this isn't C.

Refers to CLOUDDST-6236

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -84,7 +84,6 @@ def task_iib_add_bundles(
             Name of the target.
     """
     image_schema_tag = "{host}/{namespace}/{repo}:{tag}"
-    image_schema_digest = "{host}/{namespace}/{repo}@{digest}"
     verify_target_settings(target_settings)
 
     # Build new index image in IIB
@@ -93,6 +92,7 @@ def task_iib_add_bundles(
         archs=archs,
         index_image=index_image,
         deprecation_list=deprecation_list,
+        build_tags=["{0}-{1}".format(index_image.split(":")[1], task_id)],
         target_settings=target_settings,
     )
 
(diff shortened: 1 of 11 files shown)


midnightercz (approved):

pubtools/_quay/iib_operations.py
```diff
@@ -105,19 +105,19 @@ def task_iib_add_bundles(
...
     iib_namespace = build_details.index_image_resolved.split("/")[1]
-    image_digest = build_details.index_image_resolved.split("@")[1]
-    intermediate_index_image = image_schema_digest.format(
+    iib_repo = build_details.index_image_resolved.split("/")[2].split("@")[0]
+    permanent_index_image = image_schema_tag.format(
         host=target_settings.get("quay_host", "quay.io").rstrip("/"),
         namespace=iib_namespace,
-        repo="iib",
-        digest=image_digest,
+        repo=iib_repo,
```
emilyzheng, line 112: build_tags are added to intermediate index image right now. It seems an issue of IIB. I raised a question in https://code.engineering.redhat.com/gerrit/c/rcm-pub/+/289084.

querti: @emilyzheng Is this still an issue?

emilyzheng: Yeah, it's updated to get unique tag from "iib" in pub https://code.engineering.redhat.com/gerrit/c/rcm-pub/+/304487

querti: I'm not sure I understand. Is it still an issue in pubtools-quay workflow?

emilyzheng: It is. The unique tag is in repo "iib", but not in the repo parsed from index_image_resolved. So this line seems an unnecessary change.

querti: I changed the repo to "iib"

emilyzheng (approved):

PR #96, opened 2021-11-10
Skip checking for repo deprecation based on value in target settings

querti:
Until the relevant epic is completed, deprecation flag in repo
metadata should not be considered a source of truth for whether a
repo is actually deprecated. Thus, raising an error in the presence
of this flag can now be disabled in the target settings.

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -254,8 +254,9 @@ class PushDocker:
                     raise
 
             # Check if repo is not deprecated
-            # TODO: check with Comet team if this is a reliable way of checking
-            if "Deprecated" in metadata["release_categories"]:
+            if "Deprecated" in metadata["release_categories"] and target_settings.get(
+                "do_repo_deprecation_check", True
+            ):
                 raise InvalidRepository("Repository {0} is deprecated".format(repo))
 
             # if we're pushing to prod target, check if repo exists on stage as well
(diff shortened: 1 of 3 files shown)


tests/test_push_docker.py
```diff
@@ -513,6 +513,53 @@ def test_check_repos_validity_deprecated_repo(
...
+        [container_push_item_ok, container_signing_push_item],
+        hub,
+        target_settings,
+    )
+
+    mock_get_target_info.assert_called_once_with("target_stage_quay")
+    assert mock_get_repo_metadata.call_count == 3
+    mock_get_repo_metadata.call_args_list[0] == mock.call("namespace/repo1")
+    mock_get_repo_metadata.call_args_list[1] == mock.call("namespace/repo2")
+    mock_get_repo_metadata.call_args_list[1] == mock.call("namespace/repo2")
```
neithere, line 560: Duplicate comparison.

pubtools/_quay/push_docker.py
```diff
@@ -254,8 +254,9 @@ def check_repos_validity(cls, push_items, hub, target_settings):
                     raise
 
             # Check if repo is not deprecated
-            # TODO: check with Comet team if this is a reliable way of checking
-            if "Deprecated" in metadata["release_categories"]:
+            if "Deprecated" in metadata["release_categories"] and target_settings.get(
```
neithere, line 257: Same minor concern as in the related Pub review: no docs + setting name describes the action (check for X) instead of its result (stop push when X).

Feel free to ignore.

midnightercz (approved):

querti:
This change was originally made to skip checking for Deprecated flag until its implementation is ready. It turns out this was a misunderstanding and this flag is already reliable. However, we can use this change to disable checking for deprecated flag during migration, as we'll want to push old content to deprecated repos.

emilyzheng (approved):

PR #98, opened 2021-11-25
Use index_image_resolved for run_tag operation

midnightercz:


pubtools/_quay/operator_pusher.py
```diff
@@ -448,10 +448,10 @@ def push_index_images(self, iib_results, tag_suffix=None):
             _, tag = build_details.index_image.split(":", 1)
             dest_image = "{0}:{1}".format(index_image_repo, tag)
             ContainerImagePusher.run_tag_images(
-                build_details.index_image, [dest_image], True, self.target_settings
+                build_details.index_image_resolved, [dest_image], True, self.target_settings
```
querti, line 451: Do we want to do this with production index images as well? I think the justification for not using index image resolved in the past is that we always want to copy the most up-to-date image to the destination.

querti: Also, now I'm not really sure why we need intermediate index image for signing when we could have just used index image resolved. Do you remember?

querti (changes requested):

pubtools/_quay/operator_pusher.py
midnightercz: For signing we need to sign all produced index images, that's why we use intermediate index images for signing. For copying production images, yeah you're right we should copy latest

querti (approved):

PR #101, opened 2022-01-13
Allow pushing v1 containers

midnightercz:
Allow pubtools-quay to push v1 containers with same code path as pushing
source containers. By default pushing v1 containers is allowed,
it's possible to disallow this functionality via target settings
allow_v1_containers = False


midnightercz:
This change requires pub to populate pull_url for v1 containers, which is done in different change

pubtools/_quay/container_image_pusher.py
```diff
@@ -110,15 +110,19 @@ def run_tag_images(cls, source_ref, dest_refs, all_arch, target_settings):
             target_settings.get("tag_images_wait_time_increase", 10),
         )
 
-    def copy_source_push_item(self, push_item):
+    def copy_source_or_v1_push_item(self, push_item, is_source=None):
```
querti, line 113: I would prefer if you separated this into its own function. Retrofitting an existing function implies that this is a part of normal workflow, which it is not. Separating it would make it easier to remove in the future.

pubtools/_quay/container_image_pusher.py
```diff
@@ -110,15 +110,19 @@ def run_tag_images(cls, source_ref, dest_refs, all_arch, target_settings):
...
 
         Args:
             push_item (ContainerPushItem):
                 Source container push item.
         """
-        LOG.info("Copying push item '{0}' as a source image".format(push_item))
+        if is_source:
+            LOG.info("Copying push item '{0}' as a source image".format(push_item))
+        else:
+            LOG.info("Copying push item '{0}' as v1 container only".format(push_item))
```
querti, line 124: So, what would happen during signing? Would V2S2 manifest be available? We want to sign V2S2 of these images, right?

pubtools/_quay/container_image_pusher.py
```diff
@@ -266,14 +270,18 @@ def push_container_images(self):
                 .get("image", {})
                 .get("sources_for_nvr", None)
             )
-            if not sources_for_nvr and not source_ml:
+            if (
+                not sources_for_nvr
+                and not source_ml
+                and not self.target_settings.get("allow_v1_containers", True)
```
querti, line 276: It's not version 1, is it? It's version 2 schema 1

pubtools/_quay/container_image_pusher.py
```diff
@@ -266,14 +270,18 @@ def push_container_images(self):
...
+            ):
                 raise BadPushItem(
                     "Push item '{0}' contains a single-arch image that's not a "
                     "source image. This use-case is not supported".format(item)
                 )
             # Source image
-            if sources_for_nvr:
-                self.copy_source_push_item(item)
+            if sources_for_nvr and source_ml or not source_ml:
```
querti, line 283: What???????

querti (changes requested):
Could you also change the name of this PR?

pubtools/_quay/container_image_pusher.py
midnightercz: it's reported as application/vnd.docker.distribution.manifest.v1+json. But we call it v2sch1 probably because there was api with v1/ and now it's v2/ and therefore we call it v2 sch1, but docker tracks it as v1.

pubtools/_quay/container_image_pusher.py
midnightercz: it's source with manifest list or non-source without it

pubtools/_quay/container_image_pusher.py
midnightercz: good point, we should make sure there's v2sch2 digest. I guess it should be similar to situation in https://github.com/release-engineering/pubtools-quay/pull/102

pubtools/_quay/container_image_pusher.py
querti: This condition is always true

midnightercz: how's that? Its (sources_for_nvr and source_ml) or not source_ml

querti, line 283: What does it mean if metadata contain sources_for_nvr, while source has a manifest list? What kind of scenario is that? Also, source not having a manifest list doesn't conclusively make it "source or v1", it can also be non-source v2s2.

querti (changes requested):

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
+        LOG.info("{0} claim messages will be uploaded".format(len(claim_messages)))
+        signature_messages = self.get_signatures_from_radas(claim_messages)
+        self.validate_radas_messages(claim_messages, signature_messages)
+        self.upload_signatures_to_pyxis(claim_messages, signature_messages)
+        return signature_messages
+
+    @log_step("Sign container images")
+    def sign_container_images_new_digests(self, push_items):
+        """
+        Sign digests for manifest types which haven't exist before the push.
```
querti, line 527:
```suggestion
        Sign digests for manifest types which haven't existed before the push.
```

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
+            return []
+
+        LOG.info("{0} claim messages will be uploaded".format(len(claim_messages)))
+        signature_messages = self.get_signatures_from_radas(claim_messages)
+        self.validate_radas_messages(claim_messages, signature_messages)
+        self.upload_signatures_to_pyxis(claim_messages, signature_messages)
+        return signature_messages
+
+    @log_step("Sign container images")
+    def sign_container_images_new_digests(self, push_items):
```
querti, line 525: Has this been added to the docs?

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
+        """
+        Sign digests for manifest types which haven't exist before the push.
+
+        Manifest types which are not pushed in the push items can be fetched
+        from quay and then created dynamically on the fly. This methods is meant
```
querti, line 530:
```suggestion
        from quay and then created dynamically on the fly. This method is meant
```

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
+        self.validate_radas_messages(claim_messages, signature_messages)
+        self.upload_signatures_to_pyxis(claim_messages, signature_messages)
+        return signature_messages
+
+    @log_step("Sign container images")
+    def sign_container_images_new_digests(self, push_items):
+        """
+        Sign digests for manifest types which haven't exist before the push.
+
+        Manifest types which are not pushed in the push items can be fetched
```
querti, line 529: V2S1 was never meant to be supported by pubtools-quay but this solution permanently bakes it into its DNA. Future developers will wonder why it's here in the first place.

I wanted the migration related changes to be separated from the original code, explicitly noted that they exist just for the migration, and later removed in the future. You seem to disagree with this design. Can we discuss it?

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
+        from quay and then created dynamically on the fly. This methods is meant
+        to create signatures for those.
+
+        Args:
```
querti, line 533: There is also a return value

querti (changes requested):
I disagree with the design you have chosen and will be reluctant to approve it unless you present a convincing argument.

pubtools/_quay/container_image_pusher.py
```diff
@@ -266,14 +284,21 @@ def push_container_images(self):
...
-                    "Push item '{0}' contains a single-arch image that's not a "
-                    "source image. This use-case is not supported".format(item)
-                )
+                if not self.target_settings.get("allow_v1_containers", True):
+                    raise BadPushItem(
+                        "Push item '{0}' contains a single-arch image that's not a "
+                        "source image. This use-case is not supported".format(item)
+                    )
+                else:
+                    v1 = True
```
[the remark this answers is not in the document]
midnightercz: v2ch2 without manifest list can be only source

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +502,51 @@ def remove_old_signatures(
...
+        missing_digest = self.dest_quay_client.get_manifest_digest(
+            dest_ref, v2s1_manifest=v2s1_manifest
+        )
+        return missing_digest
+
+    def fetch_missing_push_items_digests(self, push_items, target_settings):
+        """Fetch digests for media types which weren't original pushed.
+
+        In order to be able to sign v2ch1 for images which were pushed as
+        v2ch2 or to sign v2ch2 for images which were pushed as v2ch1
```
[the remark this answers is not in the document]
midnightercz: This method should be universal. Disallowing pushing v1 can be simply done by disabling the target setting

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +502,51 @@ def remove_old_signatures(
...
+        fetch digests of those missing media types from quay and
+        set it to item metadata into  'new_digests' mapping
+
+        Args:
+            push_items(list): list of push items
+            target_settings(dict): target settings
+        """
+        for item in sorted(push_items):
+            item.metadata["new_digests"] = {}
```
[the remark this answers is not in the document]
midnightercz: item in this form is produced by push adapter in pub. Metadata there are just populated from push item produced by pushsource. So there's no violation of model in pushsource. Once we change the code to use push items in the form produced by pushsource directly, I think we'll have to put there some target specific field anyway

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +502,51 @@ def remove_old_signatures(
...
+                [QuayClient.MANIFEST_V2S2_TYPE, QuayClient.MANIFEST_V2S1_TYPE]
+            ) - set(item.metadata["build"]["extra"]["image"]["media_types"])
+
+            # Always add v1sch1 due to possible digest change
+            missing_media_types.add(QuayClient.MANIFEST_V2S1_TYPE)
+            for repo, tags in item.metadata["tags"].items():
+                internal_repo = get_internal_container_repo_name(repo)
+                for tag in tags:
+                    for mtype in missing_media_types:
+                        item.metadata["new_digests"][mtype] = self._fetch_digest(
```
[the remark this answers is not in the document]
midnightercz: Hmm, Right. I need to put there also tag in the mapping

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,73 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
         - send messages to RADAS and receive signatures (ManifestClaimsHandler class)
         - Upload new signatures to Pyxis
 
-        NOTE: The reason for separating the creation of schema1 and schema2 claim messages is that
-        they must happen in different stages of the push, schema2 before pushing to destination,
-        and schema1 after pushing to destination.
+        # NOTE: The reason for separating the creation of schema1 and schema2 claim messages is that
```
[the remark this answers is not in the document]
midnightercz: I forgot to remove this

pubtools/_quay/container_image_pusher.py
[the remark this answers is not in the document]
querti: Is that true? did Redhat never ship just v2s2 images similar to v2s1?

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
                     key,
                 )
 
+    def _fetch_digest(self, repo, tag, media_type):
+        """Fetch digest of repo and tag for given media type.
+
+        Args:
+            repo(str): repository name
```
querti, line 495:
```suggestion
            repo(str): Repository name.
```

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+            tag(str): image tag
```
querti, line 496:
```suggestion
            tag(str): Image tag.
```

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+            mtype(str): media type fo requested digest
+            (can be only application/vnd.docker.distribution.manifest.v2+json,
+            application/vnd.docker.distribution.manifest.v1+json)
+
+        Returns(str): manifest digest of the container
```
querti, line 501:
```suggestion
        Returns(str): Manifest digest of the container.
```

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
                     key,
                 )
 
+    def _fetch_digest(self, repo, tag, media_type):
```
querti, line 491: I'm questioning what exactly does this method accomplish. It the main purpose to combine repo + tag together into a full reference? If so, I don't believe this justifies a new method and can be done in fetch_missing_push_items_digests

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+        digest = self.dest_quay_client.get_manifest_digest(dest_ref, v2s1_manifest=v2s1_manifest)
+        return digest
+
+    def fetch_missing_push_items_digests(self, push_items, target_settings):
+        """Fetch digests for media types which weren't originally pushed.
+
+        In order to be able to sign v1 for images which were pushed as
+        v2sch2 or to sign v2sch2 for images which were pushed as v1
+        fetch digests of those missing media types from quay and
+        set it to item metadata into  'new_digests' mapping
```
querti, line 520:
```suggestion
        set it to item metadata into 'new_digests' mapping.
```

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+
+        Args:
+            push_items(list): list of push items
```
querti, line 523:
```suggestion
            push_items(list): List of push items.
```

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+            target_settings(dict): target settings
+        """
+        for item in sorted(push_items):
+            item.metadata["new_digests"] = {}
+            missing_media_types = set(
+                [QuayClient.MANIFEST_V2S2_TYPE, QuayClient.MANIFEST_V2S1_TYPE]
+            ) - set(item.metadata["build"]["extra"]["image"]["media_types"])
+
+            # Always add v1sch1 due to possible digest change
```
querti, line 532:
```suggestion
            # Always add v2s1 due to possible digest change
```

querti, line 532: Also, this comment seems rather misleading and inaccurate. We want v2s1 because we have to sign it since customers may pull it. It's unrelated to digest change.

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
+            missing_media_types.add(QuayClient.MANIFEST_V2S1_TYPE)
+            for repo, tags in item.metadata["tags"].items():
+                internal_repo = get_internal_container_repo_name(repo)
+                for tag in tags:
+                    for mtype in missing_media_types:
+                        item.metadata["new_digests"].setdefault((repo, tag), {})[
```
querti, line 538: V2S2 digest will always be the same right? Do we need to make the same query so many times? Maybe instead of looping over a set, you just always get V2S1 digests, and based on a condition get V2S2 as well (if required)?

pubtools/_quay/signature_handler.py
```diff
@@ -531,39 +492,72 @@ def sign_container_images(self, push_items, only_v2s1_manifests=False):
...
         - Upload new signatures to Pyxis
 
-        NOTE: The reason for separating the creation of schema1 and schema2 claim messages is that
-        they must happen in different stages of the push, schema2 before pushing to destination,
-        and schema1 after pushing to destination.
+        Args:
+            push_items (([ContainerPushItem])):
+                Container push items whose images will be signed.
+        Returns (list(claim-message)):
+            List of claim message sent for signing
```
querti, line 499:
```suggestion
            List of claim message sent for signing.
```

querti (changes requested):

pubtools/_quay/container_image_pusher.py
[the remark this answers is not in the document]
midnightercz: We were prepared to that eventuality - I mean in the code, but we started with manifest lists for non-source containers. I checked all the content we have and v2sch2 only tags are all marked with -source postfix

pubtools/_quay/push_docker.py
```diff
@@ -507,6 +488,57 @@ def remove_old_signatures(
...
                 )
 
+    def _fetch_digest(self, repo, tag, media_type):
+        """Fetch digest of repo and tag for given media type.
+
+        Args:
+            repo(str): repository name
+            tag(str): image tag
+            mtype(str): media type fo requested digest
+            (can be only application/vnd.docker.distribution.manifest.v2+json,
```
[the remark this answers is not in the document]
midnightercz: I think it should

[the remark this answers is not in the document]
querti: Can you check?

pubtools/_quay/container_image_pusher.py
```diff
@@ -110,20 +110,20 @@ def run_tag_images(cls, source_ref, dest_refs, all_arch, target_settings):
...
-        Perform the tagging operation for a push item containing a source image.
+    def _prepare_dest_refs(self, push_item):
+        """Prepare destination references for push.
+
+        Construct destination references based on tags and repo of push item.
 
         Args:
-            push_item (ContainerPushItem):
-                Source container push item.
+            push_item(ContainerPushItem): container push item
```
querti, line 119:
```suggestion
            push_item(ContainerPushItem): Container push item.
```

pubtools/_quay/push_docker.py
querti: What?

querti (changes requested):

pubtools/_quay/push_docker.py
midnightercz: now it's called on two places in fetch_missing_push_items_digests

pubtools/_quay/push_docker.py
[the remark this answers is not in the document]
midnightercz: fixed

PR #102, opened 2022-01-14
Sign V2S1 manifests

querti:
Since Quay.io generates V2S1 manifest for an image on demand,
signatures for digests of these manifests have to be created
as well. As the manifest values depend on the destination
location, the V2S1 manifests can only be queried after the
images are pushed. For this reason, signing of V2S1 digests
is done separately, after the image push. An unfortunate
side effect is that there will be a small time window where
a customer may pull a newly pushed image whose signature has
not been generated yet.

Multiarch images are not relevant for V2S1 because it predates
the existence of a manifest list. If the signing of V2S1 digests
fails, some newly pushed content may be left unsigned
indefinitely. I decided against implementing a rollback
functionality, since even rollback is only best-effort, and in
case of its failure nothing else guarantees the validity
of pushed data. If a push fails, the issues should be
resolved by a retried push.

The logic to remove no longer used signatures was extended for
V2S1 digests. In case of overwriting an existing tag, its
schema1 signatures will be removed as well.

A few unrelated things were fixed in this commit:
- "operator_push_items" parameter was removed from
  "remove_old_signatures" method because it wasn't used
  anywhere
- "esig" variable in "remove_old_signatures" was renamed
  to "existing_signature"
- "failed" variable in "run" is now a boolean instead of
  using 0/1 for True/False

Refers to CLOUDDST-11173

### DIFF
--- a/pubtools/_quay/push_docker.py
+++ b/pubtools/_quay/push_docker.py
@@ -25,7 +25,7 @@ LOG = logging.getLogger("pubtools.quay")
 class PushDocker:
     """Handle full Docker push workflow."""
 
-    ImageData = namedtuple("ImageData", ["repo", "tag", "digest"])
+    ImageData = namedtuple("ImageData", ["repo", "tag", "digest", "v2s1_digest"])
 
     def __init__(self, push_items, hub, task_id, target_name, target_settings):
         """
@@ -320,7 +320,7 @@ class PushDocker:
                     # repo doesn't exist, add to rollback tags
                     if not repo_tags:
                         # for rollback tags digest is not known
-                        rollback_tags.append(PushDocker.ImageData(full_repo, tag, None))
+                        rollback_tags.append(PushDocker.ImageData(full_repo, tag, None, None))
                         continue
                     # tag exists in the repo, add to backup tags
                     if tag in repo_tags.get("tags", {}):
@@ -328,8 +328,11 @@ class PushDocker:
                             host=self.quay_host, repo=full_repo, tag=tag
                         )
                         digest = self.dest_quay_client.get_manifest_digest(image_tag)
+                        v2s1_digest = self.dest_quay_client.get_manifest_digest(
+                            image_tag, v2s1_manifest=True
+                        )
                         # for backup tags store also digest
-                        image_data = PushDocker.ImageData(full_repo, tag, digest)
+                        image_data = PushDocker.ImageData(full_repo, tag, digest, v2s1_digest)
                         image = image_schema.format(
                             host=self.quay_host,
                             repo=full_repo,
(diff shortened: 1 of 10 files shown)


pubtools/_quay/signature_handler.py
```diff
@@ -488,7 +550,10 @@ def sign_container_images(self, push_items):
 
         claim_messages = []
         for item in push_items:
-            claim_messages += self.construct_item_claim_messages(item)
+            if only_v2s1_manifests:
```
midnightercz, line 553: related to my comment in pubtools/_quay/push_docker.py - can we construct here both digest everytime? Existing manifest claim will be filtered out by filter_claim_messages

midnightercz (changes requested):

pubtools/_quay/signature_handler.py
querti: What comment?

pubtools/_quay/quay_client.py
```diff
@@ -55,15 +56,36 @@ def get_manifest(self, image, raw=False, manifest_list=False):
...
+
         repo, ref = self._parse_and_validate_image_url(image)
         endpoint = "{0}/manifests/{1}".format(repo, ref)
 
+        # if V2S1 was requested, it takes precedence over the default order
+        if v2s1_manifest:
+            kwargs = {"headers": {"Accept": QuayClient.MANIFEST_V2S1_TYPE}}
+            response = self._request_quay("GET", endpoint, kwargs)
+
+            if response.headers["Content-Type"] != QuayClient.MANIFEST_V2S1_TYPE:
```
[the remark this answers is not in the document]
querti: Modified the condition.

pubtools/_quay/push_docker.py
```diff
@@ -546,7 +552,12 @@ def run(self):
         container_pusher = ContainerImagePusher(docker_push_items, self.target_settings)
         container_pusher.push_container_images()
```
midnightercz, line 554: Can we iterate over the push items here and add missing digests

pubtools/_quay/signature_handler.py
midnightercz: That's strange. I added it there now

pubtools/_quay/push_docker.py
querti: What? Add them where? And why?

midnightercz: After push_container_images we iterate over push items and based on media type of push item, other media types will be fetched form quay.io - or rather digests for manifests of those media types. With that we can push v2ch1 or v2ch2 only and when it's about to sign the manifests, we can sign both sch1 and sch2 without need to take care of what is already signed or what isn't.

querti: I think you want me to solve the issue of signing V2S2 manifests which were previously only V2S1. I think that code should be a part of #101. Signing V2S2 images after a push constitutes a special case and should not be baked into a normal workflow. I suggest you create a separate function to sign these digests and that this function will be easy to remove after migration.

midnightercz: Ok, then lets merge this one and I'll implement signing in #101

midnightercz (approved):

jimmyxiong2005 (approved):

emilyzheng (approved):

PR #104, opened 2022-01-25
Tag index image timestamps with permanent index image as a source

querti:
Two changes are made:
- For the IIB operation workflows, create a separate tagging
  operation for timestamp tags. Not including this when timestamp
  tags were introduced is an oversight, as using production tag
  as a source may reference an incorrect image (image pushed during
  separate task running in parallel). On the other hand, using
  permanent index image is safe and will ensure that every built
  image will always be available through a timestamp tag
- In the Push Docker workflow, this fact was originally accounted
  for, which is why it uses a digest to reference a source. However,
  since permanent index image is now available, it should be used
  as a source instead.

Refers to CLOUDDST-6236

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -142,7 +142,10 @@ def task_iib_add_bundles(
 
     # Push image to Quay
     ContainerImagePusher.run_tag_images(
-        build_details.index_image, [dest_image, dest_image_stamp], True, target_settings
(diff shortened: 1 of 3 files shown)


emilyzheng (commented):
There's a problem on tagging from permanent_index_image. It's started with quay.io which requires authentication in skopeo copy. Can it be changed to registry-proxy image?

emilyzheng (approved):

midnightercz (approved):

PR #135, opened 2022-06-20
Get intermediate repo from build details

emilyzheng:
The repo name is configurable in IIB, not always "iib".

Refers to CLOUDDST-13280

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -5,7 +5,12 @@ from .exceptions import InvalidTargetSettings
 from .operator_pusher import OperatorPusher
 from .signature_handler import OperatorSignatureHandler
 from .signature_remover import SignatureRemover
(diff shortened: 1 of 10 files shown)


pubtools/_quay/iib_operations.py
```diff
@@ -104,11 +104,12 @@ def task_iib_add_bundles(
         tag=tag,
     )
     # Index image used to fetch manifest list. This image will never be overwritten
-    iib_namespace = build_details.index_image_resolved.split("/")[1]
```
neithere, line 107: Looking at five nearly identical two-liners I wonder if it would make more sense to extract this snippet to a function that returns a `namedtuple`, for example.

emilyzheng: Good point! Added a function for it.

midnightercz (approved):

PR #136, opened 2022-06-21
Added support for hotfix operators

midnightercz:


pubtools/_quay/operator_pusher.py
```diff
@@ -217,7 +217,7 @@ def _get_requests_session():
         return sorted(deprecation_list)
 
     @classmethod
-    def pubtools_iib_get_common_args(cls, target_settings):
+    def pubtools_iib_get_common_args(cls, target_settings, override_settings):
```
querti, line 220: Missing docstring parameter

pubtools/_quay/operator_pusher.py
```diff
@@ -229,6 +229,12 @@ def pubtools_iib_get_common_args(cls, target_settings):
         Returns (([str]), {str:str}):
             Tuple of arguments and environment variables to be used when calling pubtools-iib.
         """
+        settings = target_settings.copy()
+        for key in ["iib_overwrite_from_index_token"]:
```
querti, line 233: What? Why is this iterated?

pubtools/_quay/operator_pusher.py
```diff
@@ -336,7 +345,7 @@ def iib_remove_operators(
                 operators, index_image
             )
         )
-        args, env_vars = cls.pubtools_iib_get_common_args(target_settings)
+        args, env_vars = cls.pubtools_iib_get_common_args(target_settings, {})
```
querti, line 348: Even if unused, override_settings should be added to iib_remove_operators as well.

pubtools/_quay/operator_pusher.py
querti: I don't think this accomplishes what you want it to accomplish

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
         for version, items in sorted(self.version_items_mapping.items()):
-            bundles = [self.public_bundle_ref(i) for i in items]
-            all_archs = [
-                i.metadata["arch"] if i.metadata["arch"] != "x86_64" else "amd64" for i in items
-            ]
-            archs = sorted(list(set(all_archs)))
-            signing_keys = sorted(list(set([item.claims_signing_key for item in items])))
+            is_hotfix = any([item.metadata.get("com.redhat.hotfix") for item in items])
+            is_advisory_source = all(
+                [re.match(r"^[A-Z0-9:\-]{4,40}$", item.origin) for item in items]
```
querti, line 434: This regex is not immediately parsable, can you add a comment on what it means?

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+            )
+            print([item.metadata.get("com.redhat.hotfix") for item in items])
```
querti, line 436: Debug prints

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                    i.metadata["arch"] if i.metadata["arch"] != "x86_64" else "amd64"
+                    for i in g_items
+                ]
+                archs = sorted(list(set(all_archs)))
+                signing_keys = sorted(list(set([item.claims_signing_key for item in items])))
 
-            # Get deprecation list
-            deprecation_list = self.get_deprecation_list(version)
+                # Get deprecation list
+                deprecation_list = self.get_deprecation_list(version)
```
querti, line 472: Isn't this redundantly called too many times? Can't it be in the outer loop?

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                        image_repo=self.target_settings["iib_index_image"], tag=tag
+                    )
+
+                bundles = [self.public_bundle_ref(i) for i in g_items]
+                all_archs = [
+                    i.metadata["arch"] if i.metadata["arch"] != "x86_64" else "amd64"
+                    for i in g_items
+                ]
+                archs = sorted(list(set(all_archs)))
+                signing_keys = sorted(list(set([item.claims_signing_key for item in items])))
```
querti, line 469: Shouldn't this be g_items as well?

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
         iib_results = {}
 
         for version, items in sorted(self.version_items_mapping.items()):
-            bundles = [self.public_bundle_ref(i) for i in items]
-            all_archs = [
-                i.metadata["arch"] if i.metadata["arch"] != "x86_64" else "amd64" for i in items
-            ]
-            archs = sorted(list(set(all_archs)))
-            signing_keys = sorted(list(set([item.claims_signing_key for item in items])))
+            is_hotfix = any([item.metadata.get("com.redhat.hotfix") for item in items])
```
querti, line 432: Is it possible that only some push items will contain this?

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                        version,
+                        items[0].metadata["com.redhat.hotfix"],
+                        items[0].origin.replace(":", "-"),
+                    )
+                    index_image = "{image_repo}:{tag}".format(
+                        image_repo=self.target_settings["iib_index_image"], tag=tag
+                    )
+                else:
+                    tag = version
+                    index_image = "{image_repo}:{tag}".format(
```
querti, line 459: Isn't this duplicate code? The image can be constructed in one place

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                    item_groups.setdefault(item.origin, []).append(item)
+            else:
+                item_groups["default"] = items
+
+            for group, g_items in item_groups.items():
+                if is_hotfix:
+                    tag = "{0}-{1}-{2}".format(
+                        version,
+                        items[0].metadata["com.redhat.hotfix"],
+                        items[0].origin.replace(":", "-"),
```
querti, line 452: Above you put items to group based on different origins, but here you consider every origin to be the same?

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                for item in items:
+                    item_groups.setdefault(item.origin, []).append(item)
+            else:
+                item_groups["default"] = items
+
+            for group, g_items in item_groups.items():
+                if is_hotfix:
+                    tag = "{0}-{1}-{2}".format(
+                        version,
+                        items[0].metadata["com.redhat.hotfix"],
```
querti, line 451: What is the value of this? Can items have different values?

pubtools/_quay/operator_pusher.py
querti: This code contacts git, so let's avoid too many calls

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
-                bundles=bundles,
-                archs=archs,
-                index_image=index_image,
-                deprecation_list=deprecation_list,
-                build_tags=["{0}-{1}".format(index_image.split(":")[1], self.task_id)],
-                target_settings=self.target_settings,
-            )
+                # build index image in IIB
+                if is_hotfix:
+                    override_settings = {"iib_overwrite_from_index": False}
```
querti, line 476: Isn't there a condition in iiblib that both or neither iib_overwrite_from_index and overwrite_from_index_token must be set? Here you only unset one.

pubtools/_quay/operator_pusher.py
```diff
@@ -420,30 +429,64 @@ def build_index_images(self):
...
+                else:
+                    override_settings = {}
+                build_details = self.iib_add_bundles(
+                    bundles=bundles,
+                    archs=archs,
+                    index_image=index_image,
+                    deprecation_list=deprecation_list,
+                    build_tags=["{0}-{1}".format(index_image.split(":")[1], self.task_id)],
+                    target_settings=self.target_settings,
+                    override_settings=override_settings,
```
querti, line 486: Wouldn't it be simpler to just copy target settings here and remove the desired parameters instead of introducing the strange new parameter "override_settings" to multiple methods?

querti (changes requested):

pubtools/_quay/operator_pusher.py
chandwanitulsi: @querti : Can you please elaborate more on your question?
FWIW, when it comes to advisories created in errata, there would be at least one  docker image that will have "com.redhat.hotfix" label and there could be other related images and builds in the same advisory that don't have the label necessarily. You can find more info in https://issues.redhat.com/browse/CWFENGINE-585.

pubtools/_quay/operator_pusher.py
midnightercz: I put it in the list as in the future there could be more override settings. Also we don't want to allow to overwrite any arbitrary settings

pubtools/_quay/operator_pusher.py
midnightercz: yeah, that's true

pubtools/_quay/operator_pusher.py
midnightercz: Original target settings shouldn't be changed. In that case you would need to copy it and do the changes on the copy. So I thought it's better to put there override

pubtools/_quay/operator_pusher.py
midnightercz: I took it from pushsource. Although I agree it's little strange

pubtools/_quay/operator_pusher.py
querti: This seems really strange to me. You give the user the option to override target settings, but in reality it's not going to work (except for one setting). The settings are already passed as a parameter, isn't it much easier and clearer to just modify them as desired before calling this method?

querti: I made it clear that I disagree with this approach, but shouldn't you also add iib_overwrite_from_index to this list?

pubtools/_quay/operator_pusher.py
querti: I think my point was, we assume that every item is hotfix if only some items (not all) contain com.redhat.hotfix. Is this correct behavior?

pubtools/_quay/operator_pusher.py
```diff
@@ -421,30 +440,60 @@ def build_index_images(self):
...
+                    i.metadata["arch"] if i.metadata["arch"] != "x86_64" else "amd64"
+                    for i in g_items
+                ]
+                archs = sorted(list(set(all_archs)))
+                signing_keys = sorted(list(set([item.claims_signing_key for item in g_items])))
+
+                # build index image in IIB
+                if is_hotfix:
+                    override_settings = {"iib_overwrite_from_index": False}
+                    override_settings["iib_overwrite_from_index_token"] = ""
```
querti, line 483: I think you could merge these two lines (if you convince me that override_settings is good design)

pubtools/_quay/operator_pusher.py
querti: Yeah, that's what I meant. Do you really think it's such a bad approach? I find override_settings very confusing. 

I mean, the function now has two parameters: a dictionary, and another dictionary to tell us how to modify the first dictionary. But the second parameter only works sometimes. I find this whole design very strange.

querti (changes requested):

pubtools/_quay/operator_pusher.py
midnightercz: yes, it's not possible to mix hotfix items with non hotfix items in one push

pubtools/_quay/operator_pusher.py
```diff
@@ -421,30 +421,60 @@ def build_index_images(self):
...
-                deprecation_list=deprecation_list,
-                build_tags=["{0}-{1}".format(index_image.split(":")[1], self.task_id)],
-                target_settings=self.target_settings,
-            )
+            for group, g_items in item_groups.items():
+                if is_hotfix:
+                    tag = "{0}-{1}-{2}".format(
+                        version,
+                        g_items[0].metadata["com.redhat.hotfix"],
+                        g_items[0].origin.replace(":", "-"),
```
emilyzheng, line 445: The format is different from what's in Pub https://code.engineering.redhat.com/gerrit/plugins/gitiles/rcm-pub/+/refs/heads/develop/pubd/pubd/push/targets/target_operators.py#422 which doesn't include advisory type.

pubtools/_quay/operator_pusher.py
```diff
@@ -421,30 +421,60 @@ def build_index_images(self):
...
+                    )
+                else:
+                    tag = version
+                index_image = "{image_repo}:{tag}".format(
+                    image_repo=self.target_settings["iib_index_image"], tag=tag
```
emilyzheng, line 450: --index-image is used as from_index in an IIB build, it should be image with version tag.

pubtools/_quay/operator_pusher.py
```diff
@@ -421,30 +421,60 @@ def build_index_images(self):
...
+                    target_settings["iib_overwrite_from_index"] = False
+                    target_settings["iib_overwrite_from_index_token"] = ""
+                else:
+                    target_settings = self.target_settings
+                build_details = self.iib_add_bundles(
+                    bundles=bundles,
+                    archs=archs,
+                    index_image=index_image,
+                    deprecation_list=deprecation_list,
+                    build_tags=["{0}-{1}".format(index_image.split(":")[1], self.task_id)],
```
emilyzheng, line 473: This is also different from hotfix implementation in Pub. Pub does IIB request with --build-tag timestamp-tag --build-tag hotfix-tag https://code.engineering.redhat.com/gerrit/plugins/gitiles/rcm-pub/+/refs/heads/develop/pubd/pubd/push/targets/target_operators.py#426. Here is --build-tag hotfix-taskid-tag.

emilyzheng (changes requested):

emilyzheng (approved):

PR #148, opened 2022-10-17
Verify bundles presence

emilyzheng:
Ensure bundles are reachable before requesting IIB. Skip building index image if timeout is reached. Skip removing old signatures if rollback happens.

Refers to CLOUDDST-15796

### DIFF
--- a/pubtools/_quay/operator_pusher.py
+++ b/pubtools/_quay/operator_pusher.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import re
 import yaml
@@ -12,6 +13,8 @@ from .utils.misc import (
     get_internal_container_repo_name,
     log_step,
     get_pyxis_ssl_paths,
+    run_with_retries,
+    get_basic_auth,
 )
 from .quay_client import QuayClient
 from .utils.misc import parse_index_image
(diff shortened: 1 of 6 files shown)


pubtools/_quay/push_docker.py
```diff
@@ -720,7 +723,7 @@ def run(self):
             if not any([x["iib_result"] for x in iib_results.values()]):
                 LOG.error("Push of all index images failed, running rollback.")
                 self.rollback(backup_tags, rollback_tags)
-                failed = True
+                sys.exit(1)
```
emilyzheng, line 726: When old tags are rolled back, their signatures shouldn't be removed.

pubtools/_quay/operator_pusher.py
```diff
@@ -395,6 +398,33 @@ def get_existing_index_images(self, quay_client):
...
+        """
+        Make sure bundles are present in Quay.
+
+        Wait until pushed images become present in Quay,
+        and return False if too much time has passed.
+        """
+        bundles = [self.public_bundle_ref(i) for i in self.push_items]
+        for bundle in bundles:
+            registry = bundle.split("/", 1)[0]
+            username, password = get_basic_auth(registry)
```
querti, line 412: How are these values added to the config file? Aren't they somewhere in the target settings? Is it possible that they wouldn't be in the config file?

emilyzheng: Bundle images are from registry(.qa|.stage).redhat.io. Auths of them are added to config.json by pub-playbook. 

Quay-mig targets have source_quay_host set to one of registry(.qa|.stage).redhat.io, and source_quay_user, source_quay_password are set with its auth. However, source_quay_host is quay.io for regular quay targets, and the username, password are quay auth.

querti (approved):

midnightercz (approved):

PR #154, opened 2023-01-23
Added support for FBC operators

midnightercz:
If repo is opted in
        If the bundle is targeted for 4.11+, do not call IIB
        If the bundle is targeted for anything until 4.10, then call IIB
        If the bundle is targeted for both SQLite and FBC (eg 4.9+), throw an error.
If the repo is not opted in
        Call iib


pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,73 @@ def build_index_images(self):
...
+        failed_items = {}
 
         for version, items in sorted(self.version_items_mapping.items()):
-            is_hotfix = any([item.metadata.get("com.redhat.hotfix") for item in items])
+            non_fbc_items = []
+            osev_tuple = tuple([int(x) for x in version.replace("v", "").split(".")])
+            for item in items:
+                for repo in item.metadata["tags"]:
+                    if repo not in repos_opted_in:
+                        repos_opted_in[repo] = pyxis_get_repo_metadata(
```
querti, line 466: It seems that the only purpose of this metadata is to throw an error if it has an unexpected value.

pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,73 @@ def build_index_images(self):
...
+                    LOG.error(
+                        "{i} is not opted in FBC but target OSE is {ose}".format(
+                            i=item, ose=version
+                        )
+                    )
+                    item.add_error(
+                        "UPLOADFAILED",
+                        "not opted in FBC but target OSE is {ose}".format(ose=version),
+                    )
+                    failed_items[id(item)] = True
```
querti, line 490: Do we still want to continue in case of this error?

pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,73 @@ def build_index_images(self):
...
             if is_hotfix and not is_advisory_source:
                 raise ValueError("Cannot push hotfixes without an advisory")
             if is_hotfix:
-                for item in items:
+                for item in non_fbc_items:
                     item_groups.setdefault(item.origin, []).append(item)
             else:
-                item_groups["default"] = items
+                item_groups["default"] = [
+                    item for item in non_fbc_items if not id(item) in failed_items
```
querti, line 514: can an item be in non_fbc_items and in failed_items at the same time?

pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,73 @@ def build_index_images(self):
...
+                    LOG.error("Inconsistent fbc opt in for item {i}".format(i=item))
+                    continue
+                elif all(item_fbc_opt_in):
+                    items_opted_in[id(item)] = True
+                else:
+                    items_opted_in[id(item)] = False
+
+                if not items_opted_in[id(item)] and osev_tuple >= (4, 11):
+                    LOG.error(
+                        "{i} is not opted in FBC but target OSE is {ose}".format(
```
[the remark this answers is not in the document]
lipoja: However for logging we should use printf-style String Formatting (`%s`), right?

pubtools/_quay/operator_pusher.py
lipoja: Just curios question: There is no pyxis API to request get metadata in bulk?

midnightercz: > Just curios question: There is no pyxis API to request get metadata in bulk?

I guess there is, but that would require now entry point in pubtools-pyxis and all together wouldn't be reviewed in time. Calling repos one by one isn't going to cause significant performance degradation

pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,97 @@ def build_index_images(self):
...
+                item.add_error(
+                    "INVALIDFILE",
+                    "Cannot push item to index image "
+                    "as it supports both <= 4.9 and >= 4.10 and is opted in FBC: {item}".format(
+                        item=item
+                    ),
+                )
+                LOG.error(
+                    "Cannot push item to index image "
+                    "as it supports both <= 4.9 and >= 4.10 and is opted in FBC: {item}".format(
```
chandwanitulsi, line 504: sqlite images are until 4.10 and fbc images start from 4.11, so this should be 
"as it supports both <= 4.10 and >= 4.11 and is opted in FBC"

tests/test_utils.py
```diff
@@ -174,3 +174,33 @@ def add_value_to_timestamp(value):
...
+
+@mock.patch("pubtools._quay.utils.misc.run_entrypoint")
+def test_get_repo_metadata(
+    mock_run_entrypoint,
+    target_settings,
+    container_multiarch_push_item,
+    operator_push_item_ok,
+    fake_cert_key_paths,
+):
+    hub = mock.MagicMock()
```
chandwanitulsi, line 187: how is mocked hub used here?

pubtools/_quay/operator_pusher.py
midnightercz: According to Yash, it was 4.9, 4.10, 4.11. Are you sure it should fail when bundle is for v4.10, v4.11 and is opted in?

pubtools/_quay/operator_pusher.py
midnightercz: yeah, that condition here is redundant

pubtools/_quay/operator_pusher.py
```diff
@@ -453,25 +453,95 @@ def build_index_images(self):
...
+        for item in self.push_items:
+            for repo in item.metadata["tags"]:
+                if repo not in repos_opted_in:
+                    repos_opted_in[repo] = pyxis_get_repo_metadata(repo, self.target_settings).get(
+                        "fbc_opt_in", False
+                    )
+            item_fbc_opt_in = [repos_opted_in[repo] for repo in item.metadata["tags"]]
+            if not (all(item_fbc_opt_in) or not any(item_fbc_opt_in)):
+                failed_items[id(item)] = True
+                item.add_error("NOTPUSHED", "Inconsistent fbc opt in")
```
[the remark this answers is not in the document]
midnightercz: What do you mean?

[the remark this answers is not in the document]
emilyzheng: I suppose push items will end up with PENDING rather than NOTPUSHED or INVALIDFILE if update_push_items is not called. The only calling I can find is https://github.com/release-engineering/pubtools-quay/blob/master/pubtools/_quay/push_docker2.py#L42, but it's obsolete code.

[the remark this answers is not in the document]
midnightercz: I believe there's pushcollector call update_push_items in pub after target operation finishes

pubtools/_quay/operator_pusher.py
midnightercz: fixed

querti (approved):

pubtools/_quay/operator_pusher.py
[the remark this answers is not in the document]
emilyzheng: Yeah, but there's sys.exit(1) after tags rollback so update_push_items in pub won't be run.

yashvardhannanavati:
@midnightercz looks good to me from FBC logic perspective :)

yashvardhannanavati (approved):

chandwanitulsi (commented):
LGTM.
Not sure if any changes are needed wrt Emily's question

PR #155, opened 2023-01-27
Drop Python2 support

lrossi-dev:
After Supercharge Pub 3 was delivered, Python 2 support is no longer required in it's dependencies as well. Dropping Python 2 will make the repos easier to maintain and will speed up Tox and Github Actions since some steps can be removed.

There are still some py2 <-> py3 compatibility workarounds that I haven't managed to remove by myself, so insight on these is more than welcome:

- [Pamiko SSH Client at command_executor.py](https://github.com/release-engineering/pubtools-quay/blob/master/pubtools/_quay/command_executor.py#L30)
- [manifest_claim encoding at signature_handler.py](https://github.com/release-engineering/pubtools-quay/blob/master/pubtools/_quay/signature_handler.py#L117)


README.rst
```diff
@@ -8,8 +8,7 @@ Set of scripts used for operating with Quay service
 Requirements
 ============
 
-* Python 2.6+
-* Python 3.5+
+* Phon 3.5+
```
midnightercz, line 11: typo

midnightercz (approved):

querti (approved):
Please make sure the CI passes

PR #161, opened 2023-02-21
Do not execute iib operation on fbc errors

midnightercz:


emilyzheng (approved):

pubtools/_quay/operator_pusher.py
```diff
@@ -506,12 +520,14 @@ def build_index_images(self):
                 )
                 failed_items[id(item)] = True
 
+        # if any of push items failed due to fbc issues, return early and skip all iib operations
+        if failed_items:
+            return {}
```
querti, line 525: Don't we want to raise instead?

midnightercz: That will be handled in push_docker. It already does check if items contain errors or not

querti (approved):

PR #175, opened 2023-06-30
Trigger building index images in parallel [CLOUDDST-18034]

zxiong:
Now, IIB has improved the capacity to be able to handle more requests at the same time, so Pub can trigger building index images in parallel.

Refers to CLOUDDST-18034.

### DIFF
--- a/pubtools/_quay/operator_pusher.py
+++ b/pubtools/_quay/operator_pusher.py
@@ -2,6 +2,9 @@ import functools
 import logging
 import re
 import yaml
+from collections import namedtuple
+from concurrent import futures
+from concurrent.futures.thread import ThreadPoolExecutor
 
 import requests
 from requests.adapters import HTTPAdapter
@@ -453,7 +456,7 @@ class OperatorPusher:
         - Use Pyxis to parse 'com.redhat.openshift.versions'
         - Get deprecation list for a given version (list of bundles to be deprecated)
         - Create mapping of which bundles should be pushed to which index image versions
-        - Contact IIB to add the bundles to the index images
+        - Contact IIB to add the bundles to the index images with multiple threads
 
         Returns ({str:dict}):
             Dictionary containing IIB results and signing keys for all OPM versions. Data will be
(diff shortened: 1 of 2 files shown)


pubtools/_quay/operator_pusher.py
```diff
@@ -524,6 +527,22 @@ def build_index_images(self):
         if failed_items:
             return {}
 
+        BuildIndexImageParam = namedtuple(
```
midnightercz, line 530: I know it's we didn't have a jira for anything like this, but since you started adding new model class here, can you please put it in new module pubtools/_quay/models.py and use dataclasses instead of namedtuple?

pubtools/_quay/operator_pusher.py
```diff
@@ -573,25 +592,52 @@ def build_index_images(self):
...
+                        target_settings=target_settings,
+                        tag=tag,
+                        signing_keys=signing_keys,
+                        is_hotfix=is_hotfix,
+                        hotfix_tag="" if not is_hotfix else hotfix_tag,
+                    )
                 )
-                iib_results[tag] = {
+
+        num_thread_build_index_images = self.target_settings.get("num_thread_build_index_images", 2)
```
midnightercz, line 617: 2 is quite low number, do you have concerns about iib performance or why you chose this number to be that low? By default number of executors for threadpool is number of cpus * 5

midnightercz (changes requested):

pubtools/_quay/operator_pusher.py
zxiong: Done.

pubtools/_quay/operator_pusher.py
zxiong: I just gave a initial value, it can be overwritten in target settings. Currently, I don't know what's the appropriate value. I will ask our Operator & Foundry team and try to estimate the value later since I think it's more depending on the capacity of IIB. Anyway, I will set the initial value to 5 since we set Pub task pod cpu.limit to 1.

midnightercz (approved):

querti (approved):

yashvardhannanavati (approved):

PR #178, opened 2023-07-25
When a command fails, print its error in multiple lines

querti:
When skopeo copy fails, the error message cannot be accessed because the log of the command output gets truncated. Since the command generates new line characters, use them to split the output and log each line separately.

Refers to CLOUDDST-19503

### DIFF
--- a/pubtools/_quay/command_executor.py
+++ b/pubtools/_quay/command_executor.py
@@ -9,6 +9,7 @@ import string
 import subprocess
 import tarfile
 import time
+import textwrap
 
 import docker
 import paramiko
@@ -182,7 +183,9 @@ class LocalExecutor(Executor):
         out, err = p.communicate(input=stdin)
 
         if p.returncode != 0 and not tolerate_err:
-            LOG.error("Command {0} failed with {1}".format(cmd, err))
+            LOG.error("Command {0} failed with the following error:".format(cmd))
+            for line in textwrap.wrap(err, 200):
+                LOG.error(f"    {line}")
             raise RuntimeError(err_msg)
 
         return out, err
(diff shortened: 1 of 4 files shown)


zxiong (approved):

pubtools/_quay/command_executor.py
```diff
@@ -182,7 +182,9 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
         out, err = p.communicate(input=stdin)
 
         if p.returncode != 0 and not tolerate_err:
-            LOG.error("Command {0} failed with {1}".format(cmd, err))
+            LOG.error("Command {0} failed with the following error:".format(cmd))
```
emilyzheng, line 185: Without this change, error can already be shown in multiple lines if it includes "\n": 

```
err = "multiline\nerror\nmessage"

2023-08-01 02:01:30 +0000 [ERROR   ] Command skopeo copy --all docker://registry-proxy.engineering.redhat.com/rh-osbs/rhel8-redis-5:rhel-8.6.0-z-containers-candidate-36737-20220620193816 docker://quay.io/redhat-dev2/test----repo1:121 --format v2s2 failed with multiline
error
message
```
Does error really have "\n"? Seems only stdout has it.

pubtools/_quay/command_executor.py
```diff
@@ -182,7 +182,9 @@ def _run_cmd(self, cmd, err_msg=None, tolerate_err=False, stdin=None):
...
+            for line in err.splitlines():
+                LOG.error(line)
```
midnightercz, line 187: Can you add an indentation here? For example "    " could be enough to make it more readable in logs

midnightercz (approved):

pubtools/_quay/command_executor.py
querti: You're right, looks like splitting by \n is not the correct solution for local executor. I have instead added textwrap that splits the long message based on length.

pubtools/_quay/command_executor.py
querti: Added.

emilyzheng (approved):

PR #180, opened 2023-07-28
Added support for pre-release containers

midnightercz:
Containers with 'com.redhat.pre-release' label are released with different tags and to different index image (if they are also operators)

### DIFF
--- a/pubtools/_quay/container_image_pusher.py
+++ b/pubtools/_quay/container_image_pusher.py
@@ -128,11 +128,16 @@ class ContainerImagePusher:
         for repo, tags in sorted(push_item.metadata["tags"].items()):
             internal_repo = get_internal_container_repo_name(repo)
             for tag in tags:
+                dst_tag = (
+                    f"{tag}.{push_item.metadata.get('com.redhat.pre-release')}"
+                    if push_item.metadata.get("com.redhat.pre-release")
+                    else f"{tag}"
+                )
                 dest_ref = image_schema.format(
                     host=self.quay_host,
                     namespace=namespace,
                     repo=internal_repo,
-                    tag=tag,
+                    tag=dst_tag,
                 )
                 dest_refs.append(dest_ref)
         return dest_refs
--- a/pubtools/_quay/container_image_pusher.py
+++ b/pubtools/_quay/container_image_pusher.py
@@ -128,16 +128,11 @@ class ContainerImagePusher:
         for repo, tags in sorted(push_item.metadata["tags"].items()):
             internal_repo = get_internal_container_repo_name(repo)
             for tag in tags:
-                dst_tag = (
-                    f"{tag}.{push_item.metadata.get('com.redhat.prerelease')}"
-                    if push_item.metadata.get("com.redhat.prerelease")
-                    else f"{tag}"
-                )
                 dest_ref = image_schema.format(
                     host=self.quay_host,
                     namespace=namespace,
                     repo=internal_repo,
-                    tag=dst_tag,
+                    tag=tag,
                 )
                 dest_refs.append(dest_ref)
         return dest_refs
--- a/pubtools/_quay/container_image_pusher.py
+++ b/pubtools/_quay/container_image_pusher.py
@@ -129,8 +129,8 @@ class ContainerImagePusher:
             internal_repo = get_internal_container_repo_name(repo)
             for tag in tags:
                 dst_tag = (
-                    f"{tag}.{push_item.metadata.get('com.redhat.pre-release')}"
-                    if push_item.metadata.get("com.redhat.pre-release")
+                    f"{tag}.{push_item.metadata.get('com.redhat.prerelease')}"
+                    if push_item.metadata.get("com.redhat.prerelease")
                     else f"{tag}"
                 )
                 dest_ref = image_schema.format(
--- a/pubtools/_quay/operator_pusher.py
+++ b/pubtools/_quay/operator_pusher.py
@@ -754,7 +754,11 @@ class OperatorPusher:
                 build_details.index_image, dest_images, True, index_image_ts
             )
             if tag_suffix:
-                dest_image = "{0}:{1}-{2}".format(index_image_repo, tag, tag_suffix)
+                _dest_images = []
+                for _dest_tag in results["destination_tags"]:
+                    _dest_images.append(
+                        "{0}:{1}-{2}".format(index_image_repo, _dest_tag, tag_suffix)
+                    )
                 ContainerImagePusher.run_tag_images(
-                    permanent_index_image, [dest_image], True, index_image_ts
+                    permanent_index_image, _dest_images, True, index_image_ts
                 )
--- a/pubtools/_quay/operator_pusher.py
+++ b/pubtools/_quay/operator_pusher.py
@@ -447,6 +447,13 @@ class OperatorPusher:
         return True
 
     def _get_fbc_opted_in_items(self):
+        """ Get items that are opted in for fbc.
+
+        An item needs to be targeted for repos with fbc_opt_in set to True and
+        ocp versions needs to be higher than 4.12. Inconsistencies in versions
+        (like support for both > 4.12 and <= 4.12) results  in item error.
+        """
+
         repos_opted_in = {}
         items_opted_in = {}
         failed_items = {}
@@ -501,7 +508,19 @@ class OperatorPusher:
                 failed_items[id(item)] = True
         return items_opted_in, failed_items
 
-    def _get_non_fbc_item_for_version(self, items, version, items_opted_in):
+    def _get_non_fbc_items_for_version(self, items, version, items_opted_in):
+        """Return non fbc items for given ocp version.
+        Args:
+            items: List[ContainerPushItem]
+                list of push items
+            version: str
+                ocp version for which items should be returned.
+            items_opted_in: Dict[int, bool]
+                list of push items opted in fbc.
+        Returns List[ContainerPushItem]:
+            list of items not opted in fbc
+        """
+
         non_fbc_items = []
         osev_tuple = tuple([int(x) for x in version.replace("v", "").split(".")])
         for item in items:
@@ -517,6 +536,20 @@ class OperatorPusher:
     def _create_item_groups_for_version(
         self, non_fbc_items, version, is_hotfix=False, is_prerelease=False
     ):
+        """Iterate thought non fbc items and group those together based on version.
+        Args:
+            non_fbc_items: List[ContainerPushItem]
+                list of items not opted in fbc
+            version: str
+                ocp version
+            is_hotfix: bool
+                flag indicating items are for hotfix push
+            is_prerelease: bool
+                flag indicating items are for prerelease push
+        Returns Dict[str, Dict[str, Any]]:
+            Dictionary of items grouped by ocp version
+        """
+
         item_groups = {
             version: {
                 "items": [],
@@ -568,7 +601,7 @@ class OperatorPusher:
             Dictionary containing IIB results and signing keys for all OPM versions. Data will be
             used in operator signing. Dictionary structure:
             {
-                "version": {
+                <target_tag>: {
                     "iib_result": (...) (object returned by iiblib)
                     "signing_keys": [...] (list of signing keys to be used for signing)
                 }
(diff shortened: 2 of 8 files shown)


pubtools/_quay/container_image_pusher.py
```diff
@@ -128,11 +128,16 @@ def _prepare_dest_refs(self, push_item):
         for repo, tags in sorted(push_item.metadata["tags"].items()):
             internal_repo = get_internal_container_repo_name(repo)
             for tag in tags:
+                dst_tag = (
+                    f"{tag}.{push_item.metadata.get('com.redhat.pre-release')}"
+                    if push_item.metadata.get("com.redhat.pre-release")
```
chandwanitulsi, line 133: Can we use "com.redhat.prerelease"  label instead? ET is also using the same to identify the bundle

pubtools/_quay/operator_pusher.py
```diff
@@ -547,39 +547,75 @@ def build_index_images(self):
...
-            item_groups = {}
+            item_groups = {
+                version: {
+                    "items": [],
+                    "overwrite": True,
+                    "build_tags": [],
+                    "destination_tags": [version],
+                }
+            }
             if is_hotfix and not is_advisory_source:
```
chandwanitulsi, line 558: cannot push hotfix or prerelease bundles without advisory

pubtools/_quay/operator_pusher.py
```diff
@@ -547,39 +547,75 @@ def build_index_images(self):
...
+                            "destination_tags": [hotfix_tag],
+                        },
+                    )
+                    item_groups[item.origin]["items"].append(item)
             else:
-                item_groups["default"] = non_fbc_items
+                for item in non_fbc_items:
+                    if item.metadata.get("com.redhat.pre-release"):
+                        item_groups.setdefault(
+                            f'{version}.{item.metadata.get("com.redhat.pre-release")}',
```
chandwanitulsi, line 582: can you please share why you are using f'{version}.{item.metadata.get("com.redhat.pre-release")}' here, but for hotfix on line 569, item.origin is used

pubtools/_quay/operator_pusher.py
```diff
@@ -547,39 +547,75 @@ def build_index_images(self):
...
+                            {
+                                "items": [],
+                                "build_tags": [
+                                    f"{version}.{item.metadata.get('com.redhat.pre-release')}"
+                                ],
+                                "overwrite": False,
+                                "destination_tags": [
```
chandwanitulsi, line 589: destination tag for pre-release bundle should also use "version-label-adv-id" format

midnightercz: ah, I see you already picked this in the document. But that would limit source for pre-releases only to advisory. Which is not necessary

pubtools/_quay/operator_pusher.py
midnightercz: because for hotfix origin is set to advisory, for pre-release content origin can be anything

pubtools/_quay/container_image_pusher.py
midnightercz: I thought it hasn't been settled yet, I don't see it in anywhere in ET code and also I haven't seen any pre-release build with that label, so this this can change I assume

querti (changes requested):
The readability of build_index_images has been significantly degraded. Can you please refactor it and split it to multiple methods?

pubtools/_quay/container_image_pusher.py
chandwanitulsi: I haven't checked ET code thoroughly, but in the testing instructions added in CWFCONF jira the label mentioned was "com.redhat.prerelease" I have linked the comment on CLOUDDST-19629

pubtools/_quay/operator_pusher.py
chandwanitulsi: I see, AFAIK pre-release content shipping is planned to be on the same lines as hotfix releases, which would involve releasing the build via the ET. 
May I ask if you can share more information or references that concludes pre-release content can be shipped via other ways?

chandwanitulsi: If 2 product teams end up using the same value for `com.redhat.pre-release` label, that would mean the old index with version-labelvalue tag would be replaced by the new index

midnightercz: hmm, good point

midnightercz:
> The readability of build_index_images has been significantly degraded. Can you please refactor it and split it to multiple methods?
Yeah, good idea

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +499,97 @@ def build_index_images(self):
...
+                    item_groups[version]["items"].append(item)
+        return item_groups
+
+    @log_step("Build index images")
+    def build_index_images(self):
+        """
+        Perform the 'build' part of the operator workflow.
+
+        This workflow is a part of push-docker operation.
+        The workflow can be summarized as:
```
querti, line 574: Can you update this description to reflect the new workflow?

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +499,97 @@ def build_index_images(self):
...
+        - Get deprecation list for a given version (list of bundles to be deprecated)
+        - Create mapping of which bundles should be pushed to which index image versions
+        - Contact IIB to add the bundles to the index images with multiple threads
+
+        Returns ({str:dict}):
+            Dictionary containing IIB results and signing keys for all OPM versions. Data will be
+            used in operator signing. Dictionary structure:
+            {
+                "version": {
+                    "iib_result": (...) (object returned by iiblib)
```
querti, line 585: This is out of date, right?

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +499,97 @@ def build_index_images(self):
...
+            if not items_opted_in[id(item)] or (items_opted_in[id(item)] and osev_tuple <= (4, 12)):
+                non_fbc_items.append(item)
+            elif items_opted_in[id(item)] and osev_tuple >= (4, 13):
+                LOG.warning(
+                    "Skipping {i}".format(i=item)
+                    + "from iib build as it's opted in for FBC and targeting OCP version >=4.13"
+                )
+        return non_fbc_items
+
+    def _create_item_groups_for_version(
```
querti, line 517: Could you add docstrings to the new methods?

querti (changes requested):

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +505,112 @@ def build_index_images(self):
...
+                    if is_hotfix
+                    else item.metadata["com.redhat.prerelease"]
+                )
+                dst_tag = "{0}-{1}-{2}".format(
+                    version,
+                    tag_part,
+                    item.origin.split("-")[1].replace(":", "-"),
+                )
+                item_groups.setdefault(
+                    item.origin,
```
querti, line 573: The description says that the items are grouped together based on version. But it looks like item_groups can only contain one version, and then it contains item.origin. What is the content of item.origin? Is it a version?

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +505,112 @@ def build_index_images(self):
...
+            is_prerelease: bool
+                flag indicating items are for prerelease push
+        Returns Dict[str, Dict[str, Any]]:
+            Dictionary of items grouped by ocp version
+        """
+        item_groups = {
+            version: {
+                "items": [],
+                "overwrite": True,
+                "build_tags": [],
```
querti, line 556: This is always empty? Later in the code when it's accessed, it seems that it's always an empty list

pubtools/_quay/operator_pusher.py
```diff
@@ -522,6 +505,112 @@ def build_index_images(self):
...
+        - Create mapping of which bundles should be pushed to which index image versions
+        - Contact IIB to add the bundles to the index images with multiple threads
+
+        Returns ({str:dict}):
+            Dictionary containing IIB results and signing keys for all OPM versions. Data will be
+            used in operator signing. Dictionary structure:
+            {
+                <target_tag>: {
+                    "iib_result": (...) (object returned by iiblib)
+                    "signing_keys": [...] (list of signing keys to be used for signing)
```
querti, line 607: This structure is not up to date

querti (changes requested):

pubtools/_quay/operator_pusher.py
midnightercz: it's used to store "{0}-{1}".format(index_image.split(":")[1], self.task_id)

querti: That's later in the code, but why is it even a part of this dictionary?

pubtools/_quay/operator_pusher.py
midnightercz: yeah, better word would be 'tag'. I'll change it

pubtools/_quay/operator_pusher.py
midnightercz: now I understand what you mean. Now it's not needed as I removed build tag for hotfix

pubtools/_quay/operator_pusher.py
```diff
@@ -623,7 +701,8 @@ def build_index_images(self):
                     "iib_result": build_details,
                     "signing_keys": param.signing_keys,
                     "is_hotfix": param.is_hotfix,
-                    "hotfix_tag": param.hotfix_tag,
+                    "destination_tags": param.destination_tags,
+                    # "hotfix_tag": param.hotfix_tag,
```
emilyzheng, line 705: Remove this? signature_handler.py also needs update to sign hotfix and prerelease images.

pubtools/_quay/operator_pusher.py
```diff
@@ -677,7 +751,7 @@ def push_index_images(self, iib_results, tag_suffix=None):
             )
 
             ContainerImagePusher.run_tag_images(
-                build_details.index_image, [dest_image], True, index_image_ts
+                build_details.index_image, dest_images, True, index_image_ts
             )
             if tag_suffix:
                 dest_image = "{0}:{1}-{2}".format(index_image_repo, tag, tag_suffix)
```
emilyzheng, line 757: Could you update the tag here as well? It's got from build_details.index_image right now.

pubtools/_quay/container_image_pusher.py
```diff
@@ -128,11 +128,16 @@ def _prepare_dest_refs(self, push_item):
         for repo, tags in sorted(push_item.metadata["tags"].items()):
             internal_repo = get_internal_container_repo_name(repo)
             for tag in tags:
+                dst_tag = (
+                    f"{tag}.{push_item.metadata.get('com.redhat.prerelease')}"
```
chandwanitulsi, line 132: The tag created provided by ET will have a 'com.redhat.prerelease' label value already, if we add it here again, the  final tag would have the label value twice, can't we use the tag from ET directly?

pubtools/_quay/operator_pusher.py
chandwanitulsi: isn't item.origin advisory-id here? so for hotfix/prerelease, the key becomes advisory-id and for others it becomes a version

Follow-up question: Do you generally add tests for this kind of private function used for breaking the workflow into multiple steps? Having a test for this would be useful

tests/test_operator_pusher.py
```diff
@@ -752,6 +744,109 @@ def test_push_operators_hotfix(
...
+    assert results == {
+        "v4.5": {
+            "iib_result": iib_results[0],
+            "signing_keys": ["some-key"],
+            "destination_tags": ["v4.5-pre-1.2-1234-4567"],
+        },
+        "v4.6": {
+            "iib_result": iib_results[1],
+            "signing_keys": ["some-key"],
+            "destination_tags": ["v4.6-pre-1.2-1234-4567"],
```
chandwanitulsi, line 803: Minor: The tag format is inconsistent here and line #782

pubtools/_quay/container_image_pusher.py
midnightercz: Yeah, I forgot they have tag template for prerelease already there

emilyzheng (approved):

PR #187, opened 2023-09-07
Unique task-id tags when pushing prerelease content

midnightercz:
Pubtools-quay call IIB to build index images with version + task-id to ensure there's preserved history of all builds requested by pub in IIB. For prerelease operators, as there's one build per operator, in the case of more operators it means same version + task-id tag is used for all of them.
That's why in the case of prerelease operators, also transformed advisory id is used

### DIFF
--- a/pubtools/_quay/operator_pusher.py
+++ b/pubtools/_quay/operator_pusher.py
@@ -655,7 +655,12 @@ class OperatorPusher:
                     image_repo=self.target_settings["iib_index_image"], tag=tag
                 )
                 build_tags = []
-                build_tags.append("{0}-{1}".format(index_image.split(":")[1], self.task_id))
+                if is_prerelease:
+                    build_tags.append(
+                        "{0}-{1}-{2}".format(index_image.split(":")[1], self.task_id, group)
+                    )
+                else:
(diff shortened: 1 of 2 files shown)


pubtools/_quay/operator_pusher.py
```diff
@@ -655,7 +655,12 @@ def build_index_images(self):
                     image_repo=self.target_settings["iib_index_image"], tag=tag
                 )
                 build_tags = []
-                build_tags.append("{0}-{1}".format(index_image.split(":")[1], self.task_id))
+                if is_prerelease:
+                    build_tags.append(
+                        "{0}-{1}-{2}".format(index_image.split(":")[1], self.task_id, group)
```
emilyzheng, line 660: group includes ":" which shouldn't be in an image tag.

emilyzheng (changes requested):

querti (commented):
It's hard to review this without proper context. Could you explain in the commit message what this change accomplishes and why it is necessary?

midnightercz:
> It's hard to review this without proper context. Could you explain in the commit message what this change accomplishes and why it is necessary?

I just added a description

querti (approved):

emilyzheng (approved):

PR #188, opened 2023-09-21
Support prerelease floating tag

midnightercz:


tests/test_operator_pusher.py
```diff
@@ -671,12 +671,12 @@ def test_push_operators_hotfix(
         IIBRes(
             "some-registry.com/index/image:5",
             "some-registry.com/index/image@sha256:a1a1",
-            ["v4.5-test-hotfix-RHBA-1234-4567"],
+            ["v4.5-test-hotfix-RHBA-1234-4567", "v4.5-test-hotfix"],
```
emilyzheng, line 674: Why adding hotfix tag in tests?

pubtools/_quay/operator_pusher.py
```diff
@@ -576,6 +576,8 @@ def _create_item_groups_for_version(
                     tag_part,
                     item.origin.split("-")[1].replace(":", "-"),
                 )
+                if is_prerelease:
```
querti, line 579: Is there a need to split it into two conditions? Can't this all be done in the second condition?

querti (changes requested):
Can you extend the commit message to explain why this change was done? I have a hard time reviewing it because I don't understand the reasoning behind it.

tests/test_operator_pusher.py
midnightercz: I added this by accident

emilyzheng (approved):

PR #189, opened 2023-09-26
Remove images created by cosign

querti:
If using an entrypoint for removing pushed images, also remove associated images created by cosign. These could be attestations, "sboms", or signatures.

Part of this PR is adding the support of OCI images to quay client and image untagger. These modules work with image manifests, so they need to understand the OCI type produced by cosign.


querti:
@emilyzheng I have addressed your comments from the previous PR.

emilyzheng (approved):

midnightercz (approved):

PR #190, opened 2023-10-02
Add --check-related-images option while calling iib-add-bundles

chandwanitulsi:
Refers to CLOUDDST-19726


chandwanitulsi:
CI failure is addressed in https://github.com/release-engineering/pubtools-quay/pull/192

querti (approved):

PR #205, opened 2023-12-07
Fix a bug where 0 IIB builds cause a push to fail

querti:
Empty iib_results is evaluated as an error and causes a rollback. However, for RHTAP pushes IIB is called 0 times and only bundle pushing is required. Fix the condition so that the push succeeds with 0 IIB builds.

Refers to CLOUDDST-21307


querti:
The change has already been tested - I will merge it without the approvals.

PR #206, opened 2023-12-07
Restore signing changes

querti:
New signing changes were not tested and had to be reverted to unblock a hotfix release of pubtools-quay. Restore the changes.


querti:
@midnightercz This PR restores the reverted signing changes back to master. Feel free to merge it when you return from your PTO.

PR #208, opened 2024-01-04
Signing refactoring

midnightercz:


pubtools/_quay/push_docker.py
```diff
@@ -556,10 +556,6 @@ def run(self):
         )
         to_sign_entries = []
         current_signatures = []
-        for item in docker_push_items:
-            to_sign_entries.extend(
-                item_processor.generate_to_sign(item, sign_only_arches=["amd64", "x86_64"])
-            )
```
emilyzheng, line 562:
```
        for to_sign_entries in to_sign_map.values():
            to_sign_entries.extend(to_sign_entries)
```
Could you also fix this? It results in incomplete and duplicate entries.

midnightercz: yeah

PR #210, opened 2024-01-10
Split signers to prepush and non prepush

midnightercz:
Cosign signer cannot work without image being available in destination registry. Therefore signer wrappers now have attribute pre_push which indicates if signer should run before container push stage and which after

### DIFF
--- a/pubtools/_quay/iib_operations.py
+++ b/pubtools/_quay/iib_operations.py
@@ -161,6 +161,7 @@ def _sign_index_image(
         signing_keys (list): List of signing keys.
         task_id (str): Task ID.
         target_settings (dict): Target settings.
+        pre_push (bool): Whether to sign before push.
     Returns:
         list: List of current signatures.
     """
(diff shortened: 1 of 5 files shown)


querti (changes requested):
What about index images? Does signing work if they are not pushed?

pubtools/_quay/iib_operations.py
```diff
@@ -169,8 +170,18 @@ def _sign_index_image(
     current_signatures: list[tuple[str, str, str]] = [
         (e.reference, e.digest, e.signing_key) for e in to_sign_entries
     ]
+    if pre_push:
```
querti, line 173: Should this be here?

querti (changes requested):

pubtools/_quay/iib_operations.py
midnightercz: you mean if it these methods shouldn't be available on the global level?

pubtools/_quay/iib_operations.py
```diff
@@ -150,6 +150,7 @@ def _sign_index_image(
     signing_keys: list[str],
     task_id: str,
     target_settings: dict[str, Any],
+    pre_push: bool = False,
```
emilyzheng, line 153: Add to docstring?

pubtools/_quay/iib_operations.py
querti: I'm trying to understand this code. My best guess: if pre_push is true, only use prepush signers, if it's false only  use pust push signers. If that's so, why not just:
SIGNER_BY_LABEL[signer["label"]].pre_push == pre_push

midnightercz: yeah, it can be simplified like that

querti (approved):

emilyzheng (approved):

PR #211, opened 2024-01-15
setup aws environment variables before signing

midnightercz:


pubtools/_quay/push_docker.py
```diff
@@ -611,6 +611,7 @@ def run(self) -> None:
             )
         # Sign containers with signers that doesn't need have pushed containers
         # in destination registry
+        set_aws_kms_environment_variables(self.target_settings, "cosign_signer")
```
querti, line 614: Should this be also added before signing prepush index images? Otherwise it will be overwritten by SBOM signing variables

pubtools/_quay/iib_operations.py
```diff
@@ -136,6 +137,7 @@ def _remove_index_image_signatures(
         current_signatures (list): List of current signatures.
         target_settings (dict): Target settings.
     """
+    set_aws_kms_environment_variables(target_settings, "cosign_signer")
```
emilyzheng, line 140: How will AWS key be used in removing signatures?

pubtools/_quay/tag_docker.py
```diff
@@ -549,6 +553,7 @@ def copy_tag_sign_images(self, push_item: Any, tag: str, executor: Executor) ->
                     continue
                 outdated_manifests.append((mad.digest, tag, repo))
 
+            set_aws_kms_environment_variables(self.target_settings, "cosign_signer")
```
emilyzheng, line 556: Is the key needed for pre-push signing?

emilyzheng (commented):
https://github.com/release-engineering/pubtools-quay/blob/master/pubtools/_quay/tag_docker.py#L655 Signing is not split here.

pubtools/_quay/iib_operations.py
midnightercz: yeah, it probably doesn't make sense to put it there as well

pubtools/_quay/tag_docker.py
midnightercz: It's not, but I thought it would be probably better to just keep setup of aws variables before every signing as it devs then doesn't need to remember and determine whether it's needed or not.

midnightercz:
> https://github.com/release-engineering/pubtools-quay/blob/master/pubtools/_quay/tag_docker.py#L655 Signing is not split here.

I'll fix it in #212

querti (approved):

pubtools/_quay/push_docker.py
```diff
@@ -661,6 +662,7 @@ def run(self) -> None:
 
             image_schema = "{host}/{namespace}/{repo}:{tag}"
 
+            set_aws_kms_environment_variables(self.target_settings, "cosign_signer")
```
emilyzheng, line 665: It's always set in _sign_index_image. I think it's not needed here.

pubtools/_quay/tag_docker.py
```diff
@@ -727,6 +733,7 @@ def untag_image(self, push_item: Any, tag: str) -> None:
 
         cert, key = get_pyxis_ssl_paths(self.target_settings)
 
+        set_aws_kms_environment_variables(self.target_settings, "cosign_signer")
```
emilyzheng, line 736: Remove signatures

pubtools/_quay/tag_docker.py
```diff
@@ -778,6 +785,7 @@ def manifest_list_remove_archs(self, push_item: Any, tag: str, remove_archs: lis
                 (to_remove_man["digest"], tag, list(push_item.repos.keys())[0])
             )
 
+        set_aws_kms_environment_variables(self.target_settings, "cosign_signer")
```
emilyzheng, line 788: Remove signatures

emilyzheng (approved):

PR #213, opened 2024-01-17
Signing fixes 2

midnightercz:
- Don't remove signatures for repushed items


pubtools/_quay/push_docker.py
```diff
@@ -731,7 +730,17 @@ def run(self) -> None:
                 failed = True
 
         # Remove old signatures
-        outdated_manifests = self.get_outdated_manifests(backup_tags)
+        # run generate backup mapping again tu push new digest of pushed containers
```
querti, line 733: Is this sentence correct?

pubtools/_quay/iib_operations.py
```diff
@@ -136,6 +137,7 @@ def _remove_index_image_signatures(
         current_signatures (list): List of current signatures.
         target_settings (dict): Target settings.
     """
+    set_aws_kms_environment_variables(target_settings, "cosign_signer")
```
querti, line 140: Another bad rebase?

pubtools/_quay/push_docker.py
```diff
@@ -731,7 +730,17 @@ def run(self) -> None:
...
+        backup_tags2, _ = self.generate_backup_mapping(docker_push_items)
+        # if new backup tag has differnet digest, it means it was overwritten during the push
```
querti, line 735: First sentence says "digest", second sentence says "manifest". Is that correct?

pubtools/_quay/push_docker.py
```diff
@@ -731,7 +730,17 @@ def run(self) -> None:
...
+        # and old signature should be removed. If manifest is the same it means, same item
+        # was just repushed
+        outdated_tags = {}
+        for bt1, bt2 in zip(backup_tags.items(), backup_tags2.items()):
+            if bt1[1] != bt2[1]:
```
querti, line 740: Is it safe to assume that the data from the two calls would correspond correctly? Wouldn't the second call of generate_backup_mapping return more backup tags - those that were previously rollback tags?

querti (changes requested):

querti (changes requested):
What does the operation key refer to?

pubtools/_quay/push_docker.py
```diff
@@ -731,7 +727,28 @@ def run(self) -> None:
...
+        outdated_tags = {}
+        backup_tags2_shared = {}
+
+        # Backup tags can contain new tags which were orignally rollback_tags
+        # limit the comparision for outdated manifests to original backup_tags only
+        for bt2 in backup_tags2.items():
+            if bt2[0] in backup_tags:
+                backup_tags2_shared[bt2[0]] = bt2[1]
+
+        for bt1, bt2 in zip(
```
querti, line 744: Are the tags guaranteed to be in the same order? Wouldn't it be simpler to only iterate backup_tags1 and check for their existence in backup_tags2 and compare the values?

querti (changes requested):

pubtools/_quay/push_docker.py
midnightercz: That's why it's iterating over backup_tags2_shared which has only tags which are both in backup_tags and backup_tags2

pubtools/_quay/push_docker.py
midnightercz: yeah, that's why backup_tags2_shared is used instead of backup_tags2

pubtools/_quay/push_docker.py
querti: Your wording implies that backup_tags2_shared can contain fewer entries than backup_tags. Can that happen? If so, we cannot rely on  the values to correspond with each other, right?

pubtools/_quay/push_docker.py
```diff
@@ -731,7 +727,28 @@ def run(self) -> None:
...
+        # and old signature should be removed. If the digest is the same it means, same item
+        # was just repushed
+        outdated_tags = {}
+        backup_tags2_shared = {}
+
+        # Backup tags can contain new tags which were orignally rollback_tags
+        # limit the comparision for outdated manifests to original backup_tags only
+        for bt2 in backup_tags2.items():
+            if bt2[0] in backup_tags:
+                backup_tags2_shared[bt2[0]] = bt2[1]
```
emilyzheng, line 742: bt2[0] is ImageData(repo=repo1, tag=tag1, v2list_digest=digest1, v2s2_digest=digest2, v2s1_digest=digest3). If digest has changed for a tag, bt2[0] will be a different ImageData and won't exist in backup_tags.

pubtools/_quay/item_processor.py
```diff
@@ -186,7 +186,6 @@ def extract_manifests(
                     results.extend(ret)
                 else:
                     results.append(ret)
-                break
```
emilyzheng, line 189: This produces duplicated to be signed amd64 manfiests from processing both manifest list and v2 manifest.

pubtools/_quay/push_docker.py
emilyzheng: backup_tags2_shared doesn't includes changed digests due to that.

pubtools/_quay/item_processor.py
```diff
@@ -181,12 +203,15 @@ def extract_manifests(
...
+                    mtype
+                ](self, image_ref, manifest, mtype, ret_headers)
+                if isinstance(mret, list):
+                    results.extend(mret)
                 else:
-                    results.append(ret)
-        return list(set(results))
+                    results.append(mret)
+        seen = set()
+        return [x for x in results if x not in seen or seen.add(x)]  # type: ignore
```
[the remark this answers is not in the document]
midnightercz: problem is that list(set(...)) doesn't preserve the order

pubtools/_quay/push_docker.py
midnightercz: yeah, good point

pubtools/_quay/clear_repo.py
```diff
@@ -25,6 +25,12 @@
         "required": True,
         "type": str,
     },
+    ("--quay-host",): {
+        "help": "Quay host name",
+        "required": True,
```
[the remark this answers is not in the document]
midnightercz: Yeah, I'll make it optional

pubtools/_quay/iib_operations.py
emilyzheng, line 1: Revert changes here?

pubtools/_quay/tag_docker.py
emilyzheng, line 1: Revert changes here?

emilyzheng (approved):

PR #227, opened 2024-02-08
Should not call IIB if bundle is opted in fbc and targets OCP >=4.11

lrossi-dev:


pubtools/_quay/operator_pusher.py
```diff
@@ -461,8 +461,8 @@ def _get_fbc_opted_in_items(self) -> tuple[dict[int, bool], dict[int, bool]]:
         """Get items that are opted in for fbc.
 
         An item needs to be targeted for repos with fbc_opt_in set to True and
-        ocp versions needs to be higher than 4.12. Inconsistencies in versions
-        (like support for both > 4.12 and <= 4.12) results  in item error.
+        ocp versions needs to be higher than 4.11. Inconsistencies in versions
```
querti, line 464: Shouldn't it be "higher than 4.10"?

lrossi-dev: Right, missed that

querti (approved):

chandwanitulsi (approved):
Could you please add #jira and description in MR?

PR #228, opened 2024-02-15
Remove old signatures for all arches not only for amd64

midnightercz:


pubtools/_quay/push_docker.py
```diff
@@ -365,68 +368,51 @@ def generate_backup_mapping(
...
-                                            else ""
-                                        ),
-                                        (
-                                            cast(ManifestArchDigest, v2s1_mad).digest
-                                            if digest_mask == 1
-                                            else ""
-                                        ),
-                                    )
-                                    backup_tags[image_data] = json.loads(mad.manifest)
+                                if mads:
```
querti, line 401: Is this condition necessary if mads is empty? Or is it here because empty list is actually [None]? Why is that the case?

pubtools/_quay/push_docker.py
```diff
@@ -365,68 +368,51 @@ def generate_backup_mapping(
...
+                                    for mad in mads:
+                                        if not mad:
```
querti, line 403: What does that mean?

pubtools/_quay/push_docker.py
```diff
@@ -365,68 +368,51 @@ def generate_backup_mapping(
...
+                                            continue
+                                        image_data = PushDocker.ImageData(
+                                            full_repo,
+                                            d_tag,
+                                            (mad.digest if digest_mask == 3 else ""),
```
querti, line 408: Maybe you can use strings instead of 1 2 3 ?

pubtools/_quay/push_docker.py
```diff
@@ -365,68 +368,51 @@ def generate_backup_mapping(
...
+                                            (mad.digest if digest_mask == 2 else ""),
+                                            (mad.digest if digest_mask == 1 else ""),
```
querti, line 410: Each mad contains manifest type param, right? Can't we use that instead of these strange loops and numbers?

querti (changes requested):

pubtools/_quay/push_docker.py
midnightercz: yeah, you're right, it's probably redundant

pubtools/_quay/push_docker.py
midnightercz: yeah, that's also possible

pubtools/_quay/push_docker.py
```diff
@@ -728,15 +728,15 @@ def run(self) -> None:
                     LOG.error("There are failed push items. Cannot continue, running rollback.")
                 else:
                     LOG.error("Push of all index images failed, running rollback.")
-                self.rollback(backup_tags, rollback_tags)
+                self.rollback(amd64_backup_tags, rollback_tags)
```
emilyzheng, line 731: Manifest list can't be restored from amd64 tags.

midnightercz: I updated amd64_backup_tags to have also manifest lists

emilyzheng: Uploading schem1 and schema2 manifests failed (errors are in testing doc). Do they need to be uploaded? I think only manifest list was uploaded before.

midnightercz: I guess it's not absolutely necessary if there's manifest list, manifests shouldn't be garbage collected so soon. But if there isn't manifest list, then those should be uploaded. I'll try to investigate why it fails

emilyzheng (approved):

querti (approved):

PR #241, opened 2024-03-06
Generate SBOM attestations for manifest lists

querti:
When attempting to get SBOMs of multiarch images referenced via tag,
cosign returns no results. The reason is that it calculates the digest
of the manifest list, which has no associated SBOM images. This can be
solved by also generating attestations for the manifest list digests.

There are no guidelines of what a ML SBOM should contain, but there is
precedent of it containing either a merge of all arch image packages, or
all arch SBOMs in a JSONL format. The second option was chosen for
implementation, which can be achieved by calling cosign attest on the
same digest multiple times.

The arch attestations that should be a part of ML attestation are
determined by inspecting the arch digests of a given ML. If a ML
attestation already exists, it's always removed and replaced by a new
one, even if there are no meaningful changes.

I believe that no changes are necessary for untagging images, since ML
digest is already added to the lost digests.

Refers to CLOUDDST-22016

### DIFF
--- a/pubtools/_quay/security_manifest_pusher.py
+++ b/pubtools/_quay/security_manifest_pusher.py
@@ -51,6 +51,7 @@ class SecurityManifestPusher:
         self.quay_host = self.target_settings.get("quay_host", "quay.io").rstrip("/")
 
         self._src_quay_client: QuayClient | None = None
+        self._dest_quay_client: QuayClient | None = None
         self._dest_quay_api_client: QuayApiClient | None = None
 
     @property
@@ -63,6 +64,17 @@ class SecurityManifestPusher:
                 self.target_settings.get("source_quay_host") or self.quay_host,
             )
         return self._src_quay_client
+
+    @property
+    def dest_quay_client(self) -> QuayClient:
+        """Create and access QuayClient for dest image."""
+        if self._dest_quay_client is None:
+            self._dest_quay_client = QuayClient(
+                self.target_settings["dest_quay_user"],
+                self.target_settings["dest_quay_password"],
+                self.quay_host,
+            )
+        return self._dest_quay_client
 
     @property
     def dest_quay_api_client(self) -> QuayApiClient:
@@ -377,8 +389,6 @@ class SecurityManifestPusher:
             destination_repos ([str]):
                 Destination paths (without a tag) where the attestation will be pushed.
             dir_path (str):
-                Directory path where artifacts will be created.
-
         """
         if item.metadata.get("product_name"):
             products = set([item.metadata["product_name"]])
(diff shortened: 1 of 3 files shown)


pubtools/_quay/security_manifest_pusher.py
```diff
@@ -429,13 +432,75 @@ def merge_and_push_security_manifest(
...
+            )
+            # Trying to determine if a multiarch attestation has changed by the new push is
+            # too complicated. Let's always remove it and replace it by a new one, even if
+            # there are no meaningful changes.
+            if attestation_exist:
+                LOG.info(
+                    f"Multiarch image {image_ref} already has an attestation. "
+                    "It will be replaced by a new attestation containing updated information."
+                )
+                self.delete_existing_attestation(image_ref, dir_path)
```
midnightercz, line 491: I wonder if it's ok to remove existing attestations. As we also don't remove existing manifests from manifest list we overwrite because they can be included in other manifest list.

querti: The old multiarch attestation is removed and replaced with a new one containing all up-to-date arch attestations. Is there a problem with this approach that I have missed?

midnightercz: My point is old attestation can be valid if a manifest in old ML is referenced by other tag

querti: That's a good point, I haven't considered ML merging.  I've changed the logic to create the ML attestations based on their arch digests. I think that should guarantee that each pushed multiarch tag should have the correct attestations.

midnightercz: I still think it won't resolve the problem. Let say you have manifest list L with manifests l1,l2,l3 and manifest list K with manifests k1, k2, l3. If you push new manifest list L2 it removes all attestation for l1,l2,l3 isn't that correct? Which means manifest list K will have manifest l3 without attestation. If I understand correctly, if you push L first and then K, l3 already has attestation and even when you recreate it at the time of push of K it will still be shared for l3 in K and L - as it's referenced by digest not by tag+digest

querti: The attestation images are not removed in such workflows, they can only be removed when explicitly deleting tags, with image_untagger.py. Image untagger makes checks so that an attestation is only removed if all the images of a given digest are removed in a repository. 

I suppose it's possible that a digest can be overwritten in a repo and an SBOM image would be left "dangling", no longer being associated with any tagged image. I don't know if it's possible in practice though, I think there are always non-floating tags being pushed which will never be overwritten.

Since the attestation images are never removed (unless explicitly deleted), I think it's safe to construct the ML attestation by gathering all its arch attestations via arch digests.

What do you think?

midnightercz: yeah, in that case it makes sense

midnightercz (approved):

PR #248, opened 2024-03-20
Filter duplicate sign entries

midnightercz:


pubtools/_quay/signer_wrapper.py
```diff
@@ -197,9 +197,14 @@ def sign_containers(
         """
         to_sign_entries = self._filter_to_sign(to_sign_entries)
         to_sign_chunks = []
+        to_sign_entries_filtered = []
```
querti, line 200: Can't you use a set?

emilyzheng (approved):

pubtools/_quay/signer_wrapper.py
midnightercz: no as SignEntry is not hashable

PR #250, opened 2024-03-27
Extract manifests from manifest list only when needed - which is when

midnightercz:
generating manifests to sign

### DIFF
--- a/pubtools/_quay/item_processor.py
+++ b/pubtools/_quay/item_processor.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 import hashlib
+import logging
 from typing import List, Dict, Any, Optional, TypeAlias, Tuple, cast, Callable
 import requests
 import time
@@ -9,6 +10,7 @@ from .quay_client import QuayClient
 from .exceptions import ManifestTypeError
 from .utils.misc import run_in_parallel, FData
 
+LOG = logging.getLogger("pubtools.quay")
 
 @dataclass
 class VirtualPushItem:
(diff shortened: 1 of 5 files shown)


pubtools/_quay/item_processor.py
```diff
@@ -79,8 +86,10 @@ def __init__(
         self.sleep_time: int = sleep_time
         self.timeout: int = timeout
         self.poll_rate: int = poll_rate
+        self.full_extract = full_extract
+        self.manifest_cache = {}
```
emilyzheng, line 90: Added log to check how manifest_cache is used:
```
diff --git a/pubtools/_quay/item_processor.py b/pubtools/_quay/item_processor.py
index 8d061e2..5879878 100644
--- a/pubtools/_quay/item_processor.py
+++ b/pubtools/_quay/item_processor.py
@@ -223,6 +223,7 @@ class ContentExtractor:
         for mtype in mtypes:
             ret_headers = {}
             if self.manifest_cache.get(image_ref, {}).get(mtype):
+                LOG.error("Get cache: %s, %s" % (image_ref, mtype))
                 manifest, ret_headers = self.manifest_cache[image_ref][mtype]
                 if manifest:
                     mret: ManifestArchDigest | List[ManifestArchDigest] = MEDIA_TYPES_PROCESS[mtype](
@@ -263,6 +264,7 @@ class ContentExtractor:
                     break
                 time.sleep(self.sleep_time)
             self.manifest_cache.setdefault(image_ref, {})[mtype] = (manifest, ret_headers)
+            LOG.error("Set cache: %s, %s" % (image_ref, mtype))
```
There's only "Set cache" but no "Get cache":
```
# curl -s http://10.0.150.153:32771/pub/task/2472/log/stdout.log | grep  "Get cache"
# curl -s http://10.0.150.153:32771/pub/task/2472/log/stdout.log | grep  "Set cache" | wc -l
11
```
So I doubt cache improved anything. Could you remove it?

emilyzheng (approved):

PR #258, opened 2024-05-20
Raise an error when manifest claims retry limit is reached

querti:
ManifestClaimsHandler automatically retries when a message encounters an error. However, when a retry limit is reached, the handler doesn't raise an error and simply continues the push with missing messages. This is obviously an issue as it results in some signatures not being created in the push. Fix the issue by raising an error when the retry limit is reached.

NOTE: At the time of writing this fix, the affected code is no longer used in production, and is superseded by pubtools-sign. The fix is implemented only if a rollback to the old code was necessary.

Refers to CLOUDDST-22737

### DIFF
--- a/pubtools/_quay/manifest_claims_handler.py
+++ b/pubtools/_quay/manifest_claims_handler.py
@@ -408,8 +408,9 @@ class _ManifestClaimsRunner(object):
         LOG.error("Error in message handler: %s", error)
 
         if self._retry_attempts > self._maximum_retries:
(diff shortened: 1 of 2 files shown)


emilyzheng:
`CODECOV_TOKEN: ` The token is empty. @midnightercz was it added to github settings as a repository secret?

emilyzheng (approved):

zxiong (approved):

PR #270, opened 2024-06-25
Added support for container identity for cosign

midnightercz:
Container identity is public facing registry + repository in public format

### DIFF
--- a/src/pubtools/_quay/push_docker.py
+++ b/src/pubtools/_quay/push_docker.py
@@ -348,6 +348,7 @@ class PushDocker:
         internal_item_processor = item_processor_for_internal_data(
             self.dest_quay_client,
             self.target_settings["quay_host"].rstrip("/"),
+            self.dest_registries,
             self.target_settings.get("retry_sleep_time", 5),
             self.target_settings["quay_namespace"],
         )
@@ -503,6 +504,7 @@ class PushDocker:
         item_processor = item_processor_for_internal_data(
             self.dest_quay_client,
             self.target_settings["quay_host"].rstrip("/"),
+            self.dest_registries,
             self.target_settings.get("retry_sleep_time", 5),
             self.target_settings["quay_namespace"],
         )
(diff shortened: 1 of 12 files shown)


src/pubtools/_quay/signer_wrapper.py
```diff
@@ -528,6 +528,20 @@ class CosignSignerWrapper(SignerWrapper):
...
+        """Return optional arguments for signing a container.
+
+        Args:
+            sign_entry (SignEntry): SignEntry to sign.
+            task_id (str): Task ID to identify the signing task if needed.
+
+        Returns:
+            dict: Optional arguments for signing a container.
+        """
+        return {"identity": sign_entry.pub_reference}
```
emilyzheng, line 543: pub_reference is a string. pubtools-sign expects identity to be a list and length of the list should be the same as digests and references.

src/pubtools/_quay/push_docker.py
```diff
@@ -557,10 +559,13 @@ def sign_new_manifests(self, docker_push_items: List[Any]) -> List[Tuple[str, st
                             + ":"
                             + tag
                         )
+                        registry = self.dest_registries[0]
+                        pub_reference = f"{registry}/{repo}"
```
emilyzheng, line 563: > The critical.identity.docker-reference field should follow the same rules / design as for the GPG signatures:
> For every tag the image is published under, there should be a signature with identity set to exactly that tag.

Does tag need to be included in identity?

src/pubtools/_quay/push_docker.py
```diff
@@ -557,10 +559,13 @@ def sign_new_manifests(self, docker_push_items: List[Any]) -> List[Tuple[str, st
                             + ":"
                             + tag
                         )
+                        registry = self.dest_registries[0]
```
emilyzheng, line 562: When an image is pulled from registry.access.redhat.com, identity in signature starts with registry.redhat.io, would signature verification fail?

src/pubtools/_quay/push_docker.py
midnightercz: Cosign doesn't include it in docker-reference. I guess it's because tag is option when you sign with cosign

src/pubtools/_quay/push_docker.py
midnightercz: I'm not exactly sure how validation works. But when doing validation in podman for example it uses containers/policy.json where you specify trusted host. SO if you have there both hosts, it should work

emilyzheng (approved):

PR #272, opened 2024-07-30
Use tag in identity refrence instead of digest

midnightercz:


src/pubtools/_quay/tag_docker.py
emilyzheng, line 1: Could you also update:
https://github.com/release-engineering/pubtools-quay/blob/master/src/pubtools/_quay/tag_docker.py#L675
https://github.com/release-engineering/pubtools-quay/blob/master/src/pubtools/_quay/tag_docker.py#L761

emilyzheng:
Does adding tag in annotations become redundant? https://github.com/release-engineering/pubtools-sign/blob/main/src/pubtools/sign/signers/cosignsigner.py#L265

midnightercz:
> Does adding tag in annotations become redundant? https://github.com/release-engineering/pubtools-sign/blob/main/src/pubtools/sign/signers/cosignsigner.py#L265

Yeah, it will. I guess we can eventually remove it

emilyzheng (approved):

src/pubtools/_quay/tag_docker.py
midnightercz: fixed

PR #273, opened 2024-08-20
Create entrypoint iib-add-deprecations [CLOUDDST-23913]

emilyzheng:
The entrypoint calls pubtools-iib-add-deprecations and then signs and published index image.

### DIFF
--- a/src/pubtools/_quay/operator_pusher.py
+++ b/src/pubtools/_quay/operator_pusher.py
@@ -392,9 +392,9 @@ class OperatorPusher:
     @tx.instrument_func()
     def iib_add_deprecations(
         cls,
-        index_image: Optional[str] = None,
-        deprecation_schema: Optional[str] = None,
-        operator_package: Optional[str] = None,
+        index_image: str,
+        deprecation_schema: str,
+        operator_package: str,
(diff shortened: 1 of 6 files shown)


midnightercz:
How do you plan to sign produced index images? Shouldn't that be also part of new ep?

midnightercz:
> > How do you plan to sign produced index images? Shouldn't that be also part of new ep?
> 
> It's signed in _sign_and_push (iib_operations.py) which is used by all iib entrypoints.

Right, I overlooked that line in the new method

midnightercz (approved):

chandwanitulsi (approved):

src/pubtools/_quay/operator_pusher.py
```diff
@@ -388,6 +388,58 @@ def iib_remove_operators(
             env_vars,
         )
 
+    @classmethod
+    @tx.instrument_func()
+    def iib_add_deprecations(
+        cls,
+        index_image: Optional[str] = None,
```
chandwanitulsi, line 395: for add_deprecations index_image should not be optional

emilyzheng: Thanks. Removed default of it.

PR #287, opened 2025-02-13
Removed manifest list merging

midnightercz:


emilyzheng:
I agree with what Tulsi commented in Jira about communicating this with potentially impacted teams.
Besides that, the change looks good to me.

PR #288, opened 2025-06-03
Changed pubtools-pyxis entry points calls

midnightercz:
Pubtools-pyxis has changed to have python entry points in different entry point group than console_scripts


midnightercz:
it should be fixed now

emilyzheng:
https://pub-hub-pub--qe.apps.int.mpp.preprod.iad2.dc.paas.redhat.com/pub/task/98448/log/stdout.log
```
2025-06-11 03:27:10 +0000 [ERROR   ] add_bundles_main() takes from 0 to 1 positional arguments but 16 were given
```

PR #301, opened 2026-07-30
Fix signer_wrapper compatibility with pubtools-sign>=1.0.0

rajulkumar:
pubtools-sign v1.0.0 changed msg_container_sign to accept signing_keys (List[str]) instead of signing_key (str) and returns "signing_keys" in its response dict instead of "signing_key". Update the call and response parsing accordingly.

Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>


src/pubtools/_quay/signer_wrapper.py
midnightercz, line 228: I think you also need to update this method to add signing_key_names in opt args
See:
https://github.com/release-engineering/pubtools-sign/blob/main/src/pubtools/sign/signers/msgsigner.py#L1088

midnightercz (changes requested):

src/pubtools/_quay/signer_wrapper.py
rajulkumar: Can you help with what value should be sent as 'signing_key_names'?
I see the values of opt args are derived from 'SignEntry' class in https://github.com/release-engineering/pubtools-quay/blob/bf444d0d37eb2bc9183955fd05bf5915912eb591/src/pubtools/_quay/item_processor.py#L34 , that doesn't have any field as signing_key_name.

rajulkumar: @midnightercz @querti ^^

midnightercz: Hey, sorry for late reply. Signing key names are the same as signing_keys, Historically those were different, then radas did some changes and now they basically only use signing_key_name

